PR #67, opened 2013-01-26
If URL is unreachable, still allow polling from other URLs in the additionalURLs list

stevewillard:
...er properties in the URL list.


rgallardo-netflix:

PR #100, opened 2013-05-11
archaius-aws: Merged contextKey and contextValue into one field

docbobo:
This pull request supersedes pull request #99. It still fixes the issue with the way the rangeKey is currently used when talking to DynamoDB. While the code can handle both the old and the new format (and in theory a mixture of both), it requires a different field to be the rangeKey (context instead of contextKey). In theory it would've been possible to reuse contextKey but I felt this was kinda ugly.

This also fixes an NPE in PropertyWithDeploymentContext#hashCode() that occurred with global properties (those having neither contextKey nor contextValue) .


rgallardo-netflix:

PR #213, opened 2015-02-05
Fix, dynamodb should retry with exponential backoff.

jakozaur:
If requests fails, retry with exponential backoff.

That prevents from cascading failures.

Best practices, according to AWS:
http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ErrorHandling.html#APIRetries

### DIFF
--- a/archaius-aws/src/main/java/com/netflix/config/sources/AbstractDynamoDbConfigurationSource.java
+++ b/archaius-aws/src/main/java/com/netflix/config/sources/AbstractDynamoDbConfigurationSource.java
@@ -56,7 +56,7 @@ public abstract class AbstractDynamoDbConfigurationSource <T> {
     static final String defaultEndpoint = "dynamodb.us-east-1.amazonaws.com";
     static final Long defaultMaxBackOffMs = 5 * 1000L;
     static final Long defaultMinBackOffMs = 500L;
-    static final Long defaultMaxRetryCount = 100;
+    static final Long defaultMaxRetryCount = 100L;
 
     //Dynamic Properties
     protected DynamicStringProperty tableName = DynamicPropertyFactory.getInstance()
@@ -113,7 +113,7 @@ public abstract class AbstractDynamoDbConfigurationSource <T> {
 
     protected ScanResult dbScanWithThroughputBackOff(ScanRequest scanRequest) {
         Long currentBackOffMs = minBackOffMs.get();
-        Long retryCount = 0;
+        Long retryCount = 0L;
         while (true) {
             try {
                 return dbClient.scan(scanRequest);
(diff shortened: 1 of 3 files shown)


archaius-aws/src/main/java/com/netflix/config/sources/AbstractDynamoDbConfigurationSource.java
```diff
@@ -41,12 +45,16 @@
     static final String keyAttributePropertyName = "com.netflix.config.dynamo.keyAttributeName";
     static final String valueAttributePropertyName = "com.netflix.config.dynamo.valueAttributeName";
     static final String endpointPropertyName = "com.netflix.config.dynamo.endpoint";
+    static final String pollingMaxBackOff = "com.netflix.config.dynamo.maxPollingBackOff";
```
gorzell: It would be great if you identified the time unit (i.e. ms) in the variable name, or at the very least put in a comment.  I am assuming it is milliseconds.

archaius-aws/src/main/java/com/netflix/config/sources/AbstractDynamoDbConfigurationSource.java
```diff
@@ -41,12 +45,16 @@
...
+    static final String pollingMinBackOff = "com.netflix.config.dynamo.maxPollingBackOff";
 
     //Property defaults
     static final String defaultTable = "archaiusProperties";
     static final String defaultKeyAttribute = "key";
     static final String defaultValueAttribute = "value";
     static final String defaultEndpoint = "dynamodb.us-east-1.amazonaws.com";
+    static final Long defaultMaxBackOff = 5 * 1000L;
```
gorzell: See above.

archaius-aws/src/main/java/com/netflix/config/sources/AbstractDynamoDbConfigurationSource.java
gorzell: I would also clarify that this is the property name not the property property object.  Notice how the other names are clearly labeled as PropertyName.

archaius-aws/src/main/java/com/netflix/config/sources/AbstractDynamoDbConfigurationSource.java
```diff
@@ -95,6 +106,25 @@ public AbstractDynamoDbConfigurationSource(AmazonDynamoDB dbClient) {
...
+
+    protected ScanResult dbScanWithThroughputBackOff(ScanRequest scanRequest) {
+        Long currentBackOff = minBackOff.get();
+        while(true) {
+            try {
+                return dbClient.scan(scanRequest);
+            }
+            catch (ProvisionedThroughputExceededException e) {
+                currentBackOff = Math.min(currentBackOff * 2, maxBackOff.get());
+                log.info("Failed to poll Dynamo due to ProvisionedThroughputExceededException. Backing off.");
```
gorzell: Since this is an infinite loop I think at least a WARN if not an ERROR is justified here.

gorzell: I would also put the currentBackOff value in the log string.

gorzell:
Looks good other than some minor fixes.  I worry a little bit about the infinite loop and what thread might get blocked.

Also would be nice if there were some tests, but since it relies on an interaction with the AWS client am a not sure there is a simple way to do that.

jakozaur:
Implemented review fixes. Also wonders about infinite loop. Maybe there should be some max ups.

For my use case it's fine, we got different part which increase DynamoDB throughput. We just need to be able temporary hold it gracefully. However, that might be not true for other users of this library.

Let me know if you would like me to do something about it.

gorzell:
Why don't you put in a max retries just to be safe.  You can use -1 as infinite.

PR #215, opened 2015-02-28
Archaius Version 2.0

elandau:
WORK IN PROGRESS

This is the initial implementation of Archaius v2 without a dependency on apache configuration.  This implementation is far simpler, has no static state and uses Typesafe Config for loading of external resources.   

This code is not yet at feature parity with 1.0 but is at a state where we can start discussing the merits of this approach.  In addition we will be creating an archaius-legacy subproject with bindings to the older static API.


elandau:
This PR is replaced by #216 which uses a proper clean branch

PR #216, opened 2015-02-28
Archaius 2.x

elandau:
WORK IN PROGRESS

This is the initial implementation of Archaius v2 without a dependency on apache configuration. This implementation is far simpler, has no static state and uses Typesafe Config for loading of external resources.

This code is not yet at feature parity with 1.0 but is at a state where we can start discussing the merits of this approach. In addition we will be creating an archaius-legacy subproject with bindings to the older static API.

### DIFF
--- a/archaius-core/src/main/java/netflix/archaius/CascadeStrategy.java
+++ b/archaius-core/src/main/java/netflix/archaius/CascadeStrategy.java
@@ -1,0 +1,29 @@
+package netflix.archaius;
+
+import java.util.List;
+
+/**
+ * Strategy for determining a set of cascading resource names.  The strategy will resolve
+ * a single resource name into an ordered list of alternative names.
+ * 
+ * For example, a strategy may specify that additional configuration files may be loaded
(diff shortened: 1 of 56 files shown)


archaius-core/src/main/java/netflix/archaius/CascadeStrategy.java
```diff
@@ -0,0 +1,29 @@
+package netflix.archaius;
```
spencergibb: Curious why the new package structure?  Why drop the `com.`?

spencergibb:
:+1: looking forward to this.

archaius-core/src/main/java/netflix/archaius/CascadeStrategy.java
elandau: There's a growing trend to remove the com. as it is completely unnecessary.  :)

gorzell: It's only completely unnecessary if you have a monopoly on all of the possible domains. ;-)

PR #218, opened 2015-03-01
Remove usage of Guava method removed in future versions

jsmodic:
This commit removes usage of Closeables#closeQuietly(Closeable).
It is removed in future versions of Guava and prevents Archaius
from otherwise being able to run with the latest Guava.

It keeps the same behavior as before and logs the same statement.

### DIFF
--- a/archaius-zookeeper/src/main/java/com/netflix/config/source/ZooKeeperConfigurationSource.java
+++ b/archaius-zookeeper/src/main/java/com/netflix/config/source/ZooKeeperConfigurationSource.java
@@ -227,7 +227,7 @@ public class ZooKeeperConfigurationSource implements WatchedConfigurationSource,
         try {
             Closeables.close(pathChildrenCache, true);
         } catch (IOException exc) {
-            logger.warn("IOException should not have been thrown.", exc);
+            logger.error("IOException should not have been thrown.", exc);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


archaius-zookeeper/src/main/java/com/netflix/config/source/ZooKeeperConfigurationSource.java
```diff
@@ -223,6 +224,10 @@ synchronized void deleteZkProperty(String key) throws Exception {
     }
     
     public void close() {
-    	Closeables.closeQuietly(pathChildrenCache);
+        try {
+            Closeables.close(pathChildrenCache, true);
+        } catch (IOException exc) {
+            logger.warn("IOException should not have been thrown.", exc);
```
gorzell: If you want to maintain the same behavior as before you should probably log this as ERROR since guava is logging it as SEVERE.

jsmodic: That's reasonable. I'll update this patch.

Though I think the same behavior is inherently altered as it's going to SLF4J instead of JUL (and I'd be surprised if most users were importing a JUL-SLF4J bridge). But I'm sure nobody wants more JUL!

gorzell: Thanks for the fix.  Happy to merge it when I see the new patch come by.

PR #237, opened 2015-03-06
add Decoder interface to allow plugability of decoding properties to arb...

spencergibb:
...itrary values.

### DIFF
--- a/archaius-core/src/main/java/netflix/archaius/Decoder.java
+++ b/archaius-core/src/main/java/netflix/archaius/Decoder.java
@@ -1,0 +1,11 @@
+package netflix.archaius;
+
+/**
+ * API for decoding properties to arbitrary types.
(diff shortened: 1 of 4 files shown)


spencergibb:
build failed because the `gradle-wrapper.jar` is not checked in.

elandau:
Thanks for the PR!  Glad to see others getting involved.  Do you see this being extended to include parsing of all types, getLong, getInt, etc..  I'm currently doing that in AbstractConfig but feel it should be implemented elsewhere.

spencergibb:
@elandau Yeah, wouldn't be hard to move to the long, int parsing to the decoder.  Spring has robust conversion framework that would get plugged in [spring-cloud-netflix](https://github.com/spring-cloud/spring-cloud-netflix).

PR #239, opened 2015-03-06
Fix the build so that it will actually work.

gorzell:
This also meant commenting out the ignore line for *.jar, which I am not sure is appropriate anyway, but happy to discuss if people think there is  a good reason for it to be there.


gorzell:
Well it was a fun try, but it turns out this also depends on JDK7, which the build isn't setup to use.  Is that an intentional change?  If so why not just go all the way to 8?

```
archaius-core/src/main/java/netflix/archaius/config/EmptyConfig.java:24: cannot find symbol
symbol  : method emptyIterator()
location: class java.util.Collections
        return Collections.emptyIterator();
```

gorzell:
Probably not surprisingly a lot of the test appear to be broken.  But with this PR and a JDK change on the build job everything should at least compile.  I will probably try and clean up the build file a bit more given some of the generalities it contains.

elandau:
Build for 2.x succeeds as of #242

gorzell:
Sweet, what did you do about the JVM version?

elandau:
Modified the cloudbees configuration to java7.  We can't go to java8 yet because most of netflix libraries are still on java7.

gorzell:
That should be fine, but there is a consequence that you are basically
forcing 1.x on to java 7 as well.

PR #240, opened 2015-03-07
Add build.gradle to each subproject

elandau:


gorzell:
Glad you did this, it was next on my list, but I wasn't sure how people would feel about it.  I like this style a lot more than one giant build file.

PR #255, opened 2015-03-10
Start to segregate property conversions from ObservableProperty.

gorzell:


archaius-core/src/main/java/netflix/archaius/ObservableProperty.java
```diff
@@ -1,8 +1,5 @@
 package netflix.archaius;
 
-import java.math.BigDecimal;
-import java.math.BigInteger;
-
 /**
  * Create an observable stream of property value changes.  When subscribed to
```
gorzell: I am not sure that this comment/description is still accurate based on the contents of the interface.

PR #256, opened 2015-03-10
Fixes and updates to the documentation

gorzell:
I'll probably re-use this branch for doc changes in general.  It currently depends on/includes the changes in #255.


archaius-core/src/main/java/netflix/archaius/PropertyConversions.java
```diff
@@ -0,0 +1,66 @@
+package netflix.archaius;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+
+/**
+ * Interface for converting properties between various types.
+ */
+public interface PropertyConversions {
```
elandau: Why do you think we need this abstraction?  I believe the Decoder is better suited for this separation.  The ObservableProperty is more a proxy to making the correct getXXX on Config which in turn should call the Decoder.

gorzell: Having this functionality in ObservableProperty feels conflated to me, as it has basically nothing to with a property being Observable.  Another option would be to put it directly in the Property interface but that seemed like it might be overly strict.

archaius-core/src/main/java/netflix/archaius/ObservableProperty.java
```diff
@@ -1,8 +1,5 @@
 package netflix.archaius;
 
-import java.math.BigDecimal;
-import java.math.BigInteger;
-
 /**
  * Create an observable stream of property value changes.  When subscribed to
  * the ObservableProperty will emit either the current value or default if no
```
gorzell: I'm not sure that this javadoc is still accurate.

PR #261, opened 2015-03-11
Move all packages to com.netflix.archaius (from netflix.archaius)

elandau:


spencergibb:
Since I asked earlier about the package name, why the change back to `com.`?

elandau:
The move back to com. had to do with making archaius 2.x org and package names compatible with package names in 0.x.   I still like the idea of dropping the com. but I shouldn't have attempted this on a 2.x of an existing project.

spencergibb:
:-)

elandau:
@spencergibb Can you do a thorough review of the code and API.  I actually built an archaius 2.x rc.1 so people can start trying it out but there's still time to make API changes.

spencergibb:
@elandau will do.  It may be tomorrow or monday.

PR #265, opened 2015-03-12
Fix handling of parse and missing value errors

elandau:


archaius-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -201,130 +201,130 @@ public Short getShort(String key) {
...
     }
 
     @Override
     public BigInteger getBigInteger(String key) {
-        String value = getString(key);
+        String value = getRawString(key);
         if (value == null) 
             return notFound();
         try {
             return BigInteger.valueOf(Long.valueOf(value));
```
brharrington: Not related to the specific change in this PR, but shouldn't this do something like:

```java
return new BigInteger(value);
```

So the value doesn't have to fit within a Long?

archaius-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -201,130 +201,130 @@ public Short getShort(String key) {
...
     }
 
     @Override
     public BigDecimal getBigDecimal(String key) {
-        String value = getString(key);
+        String value = getRawString(key);
         if (value == null) 
             return notFound();
         try {
             return BigDecimal.valueOf(Long.valueOf(value));
```
brharrington: Same as with `getBigInteger`, but it seems really strange here to force a conversion to Long first.

archaius-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
elandau: yes.  good catch.

PR #266, opened 2015-03-12
test case for typesafe config mapping

brharrington:
I was curious about how some constructs would work between
an implementation based on typesafe or commons-configuration.
For example, both support a notion of lists, but it looks like
archaius is always expecting an underlying structure that is
essentially `Map<String, String>`.

In this test case getList fails completely if a list is used
in the HOCON format. For another example, a duration can be
retrieved as a string, but not as a numeric value or duration.

What is the expected behavior? I thought the reason
`getRawProperty` returned `Object` was to have more flexibility
in passing these sort of types from an implementation like
typesafe config or commons-configuration, but #265 changes
it to force a string.

### DIFF
--- a/archaius-typesafe/src/test/java/com/netflix/archaius/typesafe/TypesafeConfigLoaderTest.java
+++ b/archaius-typesafe/src/test/java/com/netflix/archaius/typesafe/TypesafeConfigLoaderTest.java
@@ -8,7 +8,6 @@ import com.netflix.archaius.DefaultConfigLoader;
 import com.netflix.archaius.cascade.ConcatCascadeStrategy;
 import com.netflix.archaius.config.MapConfig;
 import com.netflix.archaius.exceptions.ConfigException;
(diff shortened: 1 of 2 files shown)


elandau:
First, archaius was always doing a toString() so the previous method signature was very misleading.  Also, archaius historically dealt with lists and a single property with delimited values. I'd like to maintain the previous behavior here.

With regards to getting Archaius to work with Typesafe's Lists here's one approach,
1.  getRawString() should be removed from the Config interface and placed in AbstractConfig.  Archaius internal configurations can continue to treat lists as parsed single properties
2.  CascadingCompositeConfig needs to be modified to delegate all parsing of values to the first child that has the property so it doesn't make any assumptions regarding how the values are stored.
3.  TypesafeConfig will implement Config directly and return the values as parsed and interpolated by Typesafe Config's implementation.  The tricky part will be to incorporate the string interpolator so that the Tyepsafe Config will be able to replace with property values from archaius.

PR #276, opened 2015-03-20
fix javadoc for specifying defaults with Property

brharrington:
I find having to specify the default for every access a bit
cumbersome. In general it means that you have to have a constant
somewhere for the default and pass it in on every access.

I tend to favor the DRY principle in regards to specifying defaults,
more similar to the reasoning in:

https://github.com/typesafehub/config#how-to-handle-defaults

Further, forcing properties to be specified as part of the config
files is quite useful for debugging because it means the value you
see in views like the karyon admin are the full set of properties
used. With the defaults all over the place you essentially have to
read the code if a value is missing to find out what value would
get used.

### DIFF
--- a/archaius-core/src/main/java/com/netflix/archaius/Property.java
+++ b/archaius-core/src/main/java/com/netflix/archaius/Property.java
@@ -27,12 +27,12 @@ import java.util.concurrent.TimeUnit;
  *     private final Property<String> prop;
  *     
  *     MyService(PropertyFactroy config) {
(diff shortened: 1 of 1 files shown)


elandau:
Thanks for the fix, but... after some offline discussions I'm actually going to return the defaults to the way they are described in the documentation here.

brharrington:
Ok, that is a preferable from an API standpoint for me, so I'll close this out. I though you mentioned in one of the other threads that caused some caching issues though? Did you find another workaround for that?

PR #277, opened 2015-03-20
Major cleanup

elandau:
AppConfig API more explicit about adding Configs to library and override layer
Consolidate DynamicConfig API into Config.  Consolidate all listener interfaces into ConfigListener.
Rename Config#containsProperty to Config#containsKey to be consistent with other method names
Renamed Config#subset to Config#getPrefixedView
Property now has an inherent default instead of passing default as an argument to get
Remove CompositeConfig#addConfigLast and CompositeConfig#addConfigFirst and replace with a single CompositeConfig#addConfig.


archaius-core/src/main/java/com/netflix/archaius/Config.java
```diff
@@ -66,18 +98,47 @@
...
+     */
+    boolean containsKey(String key);
     
+    /**
+     * @return True if empty or false otherwise.
+     */
     boolean isEmpty();
     
+    /**
+     * Return an interator to all property names owned by this conifg
```
brharrington: s/interator/iterator/

archaius-core/src/main/java/com/netflix/archaius/ConfigListener.java
```diff
@@ -22,14 +22,28 @@
...
+    public void onConfigRemoved(Config config);
+    
     /**
      * Notify the parent that the value of a key has changed.  The parent is expected
      * to refresh its complete state.
      * 
      * @param propName
      */
-    public void onUpdate(String propName, Config config);
+    public void onConfigUpdated(String propName, Config config);
```
brharrington: For a given update is it guaranteed that only one of the `onConfigUpdated` variants will get called? Is this variant really needed? It seems like listener implementations would need to implement the other variant and not expect to get the `propName` explicitly anyway.

archaius-core/src/main/java/com/netflix/archaius/config/InterpolatingConfig.java
```diff
@@ -0,0 +1,347 @@
...
+        if (value == null) {
+            return notFound(key, defaultValue);
+        }
+        String[] parts = value.split(",");
+        return Arrays.asList(parts);
+    }
+
+    @Override
+    public <T> T get(Class<T> type, String key) {
+        return get(type, key, null);
```
brharrington: This method behaves differently than the others if there is no value for they key. The others with no default throw `NoSuchElementException`, this one is treating it as a default of `null`.

archaius-core/src/main/java/com/netflix/archaius/config/InterpolatingConfig.java
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    @Override
+    public <T> T get(Class<T> type, String key, T defaultValue) {
```
brharrington: Is there any reason the other getXXX methods cannot delegate to the one for the generic class? For example:

```java
@Override
public Short getShort(String key, Short defaultValue) {
    return get(Short.class, key, defaultValue);
}
```

settings.gradle
```diff
@@ -2,5 +2,5 @@ rootProject.name='archaius'
 include 'archaius-core'
 include 'archaius-typesafe'
 include 'archaius-commons-configuration'
-include 'archaius-legacy'
+//include 'archaius-legacy'
```
brharrington: Is this intentional?

archaius-core/src/main/java/com/netflix/archaius/ConfigListener.java
elandau: If it's a snapshot update then on onConfigUpdated(Config config) will be called.  It it's an incremental update then onConfigUpdated(String propName, Config config) will be called multiple times but onConfigUpdated(Config config) will not be called.

archaius-core/src/main/java/com/netflix/archaius/config/InterpolatingConfig.java
elandau: Fixed in #278

settings.gradle
elandau: no, fixed in #278

archaius-core/src/main/java/com/netflix/archaius/config/InterpolatingConfig.java
elandau: Will need to give this some thought.

archaius-core/src/main/java/com/netflix/archaius/Property.java
```diff
@@ -48,7 +48,7 @@
     /**
      * @return  Most recent value for the property
      */
-    T get(T defaultValue);
+    T get();
```
gorzell: It might be worth also having a getOrElse(T else) method so that you could override the general default for a specific call.

PR #279, opened 2015-03-24
cached properties need to refresh

brharrington:
The CachedProperty was not getting updated when changes
were made to the config.


elandau:
Better fix in PR #281

PR #284, opened 2015-03-27
Fix configuration mapping when there is no prefix

elandau:
Additional cleanup
* Argument validation in DefaultAppConfig
* Move annotations to a better package


archaius-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
```diff
@@ -123,7 +127,9 @@ public Builder withOverrideLayer(boolean flag) {
          * will use PropertiesConfigLoader.
          */
         public Builder withConfigReader(ConfigReader loader) {
-            this.loaders.add(loader);
+            if (loader != null) {   
+                this.loaders.add(loader);
```
brharrington: Why does this method ignore null values where the others in this builder throw?

PR #287, opened 2015-03-30
fix #286, easier to customize AppConfig

### LINKED ISSUE #286: 2.x make it easier to customize AppConfig
The current guice module for archaius has protected methods with `@Provides` for creating the `AppConfig`:

https://github.com/Netflix/archaius/blob/2.x/archaius-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java

In guice4 overriding an `@Provides` method is not allowed, see the following discussion for the rationale:

https://groups.google.com/forum/m/#!topic/google-guice/bRo5SvmzpdI

A user would have to use `Modules.override` to customize the `AppConfig` which doesn't seem like it should be the case as it is a pretty common use-case. I think it will also end up making upgrading guice more painful down the road because it encourages creating a sub-class in a way that won't work with guice4.

brharrington:
This change splits out the two `@Provides` methods for the
`ConfigMapper` and `AppConfig` into a `DefaultArchaiusModule`.
The base `ArchaiusModule` expects something else to provide
those bindings.

The other advantage to this over expecting a sub-class to
override is that the user is free to inject whatever they need
into their `@Provides` for the `AppConfig`.

README has been updated to show the usage for both `ArchaiusModule`
and `DefaultArchaiusModule`.

### DIFF
--- a/archaius-guice/README.md
+++ b/archaius-guice/README.md
@@ -9,12 +9,24 @@ To enable create Guice bindings for Config and AppConfig and install a single gu
 
 ```java
 Injector injector = Guice.createInjector(
+    new DefaultArchaiusModule()
+    );
+```
+
+For users that need to customize the `AppConfig` object, the `AchaiusModule` can be used
+instead. The user is responsible for providing bindings for the `ConfigMapper` and `AppConfig`
+instances to use.
+
+```java
+Injector injector = Guice.createInjector(
     new AbstractModule() {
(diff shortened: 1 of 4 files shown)


elandau:
We should avoid inheritance with ArchaiusModule and just allow for an externally provided AppConfig.  

See https://github.com/Netflix/archaius/pull/295

PR #295, opened 2015-04-13
Cleanup ArchaiusModule

elandau:
Simplify bindings in ArchaiusModule and support an externally provided A
ppConfig implementation instead of allowing ArcahiusModule to be inherited.

### DIFF
--- a/archaius-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
+++ b/archaius-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
@@ -16,12 +16,12 @@
 package com.netflix.archaius.guice;
 
 import javax.inject.Inject;
-import javax.inject.Provider;
 import javax.inject.Singleton;
 
 import com.google.inject.AbstractModule;
 import com.google.inject.Injector;
 import com.google.inject.Key;
+import com.google.inject.Provider;
 import com.google.inject.Provides;
 import com.google.inject.ProvisionException;
 import com.google.inject.Scopes;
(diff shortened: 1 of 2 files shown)


brharrington:
This approach will still require using `Module.override` if something needs to be injected to create the initial AppConfig object. For example, say I wanted to inject `DiscoveryClient` to lookup the set of platformservice instances.

archaius-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -139,37 +138,39 @@ public void afterInjection(I injectee) {
...
+    
+    public ArchaiusModule() {
+        this(null);
+    }
+    
+    public ArchaiusModule(AppConfig appConfig) {
+        this.appConfig = appConfig;
+    }
+    
+    public static ArchaiusModule fromAppConfig(final AppConfig config) {
```
brharrington: What is the rationale having the public constructor that takes an `AppConfig` and this method?

archaius-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -139,37 +138,39 @@ public void afterInjection(I injectee) {
...
         requestInjection(listener);
-        
         bindListener(Matchers.any(), listener);
+        
+        bind(ConfigMapper.class).to(DefaultConfigMapper.class).in(Scopes.SINGLETON);
     }
     
     @Provides
     @Singleton
-    protected ConfigMapper createConfigMapper() {
```
brharrington: I haven't needed to do it yet, but I'm curious, how often is it expected that the user will  need to customize the `ConfigMapper`?

elandau:
I modified the Module to force the use of Modules.override in order to provide a different implementation of AppConfig.  This is actually a better use of Guice Modules since modules really shouldn't have any state and passing AppConfig to a module violates that principle.

archaius-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
elandau: I don't expect a need for a different ConfigMapper implementation.

PR #296, opened 2015-04-17
Rename artifacts to archaius2

elandau:
Rename artifacts to archaius2 to avoid conflicts with archaius 1 so they can be loaded together during migration


gorzell:
This seems a bit like a hack.  I realize it's a substantial change, but putting versions in the jar name seems extreme. How common is the case of needing to load both?

elandau:
I agree that this is not ideal but is better than any alternative that I can think of.  At Netflix we have a large codebase of libraries and applications and it would be very difficult to replace the code backing the old API with the new API without some changes in expected behavior.  Also, since the old API is very static heavy it does not fit well with the new API which promotes DI best practices.  As a result we are opting to run both configuration versions in parallel and allow application and libraries to migrate code to the new API at their own pace.

NiteshKant:
>> How common is the case of needing to load both?

Very very common, unless it is a greenfield development project. Archaius is used in libraries and these libraries use different versions of archaius. If we can only tolerate a single version of archaius in an application, no library with a new archiaus version can be released till every other library move to the new  version, which quiet understandably will be a herculean task :)

Having different module names (appending the version) is the path of least resistance from the point of view of build and publish and we are following this model in [karyon](https://github.com/Netflix/karyon), [eureka](https://github.com/Netflix/eureka/tree/2.x) and any other project that is changing major versions with incompatible APIs.

PR #297, opened 2015-04-25
DefaultAppConfig simplification

elandau:
Modify DefaultAppConfig to be a DelegatingConfig instead of a CascadingCompositeConfig to avoid exposing all of the CascadingCompositeConfig functionality.
Remove withLoadSystem from ConfigLoader
Expose getting any layer from AppConfig

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
@@ -16,6 +16,8 @@
 package com.netflix.archaius;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Properties;
 
@@ -24,6 +26,8 @@ import org.slf4j.LoggerFactory;
 
 import com.netflix.archaius.cascade.SimpleCascadeStrategy;
 import com.netflix.archaius.config.CascadingCompositeConfig;
+import com.netflix.archaius.config.CompositeConfig;
+import com.netflix.archaius.config.DelegatingConfig;
 import com.netflix.archaius.config.EnvironmentConfig;
 import com.netflix.archaius.config.SimpleDynamicConfig;
 import com.netflix.archaius.config.SystemConfig;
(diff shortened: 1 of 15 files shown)


archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
```diff
@@ -205,46 +182,40 @@ public static Builder builder() {
         return new Builder();
     }
     
-    public static DefaultAppConfig createDefault() {
+    public static DefaultAppConfig createDefault() throws ConfigException {
         return new Builder().build();
     }
     
+    private final CascadingCompositeConfig     root;
     private final SimpleDynamicConfig          runtime;
```
brharrington: If I'm reading this correctly this is now the override layer. Can we name the variable override so it is easier to follow the code?

archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
```diff
@@ -205,46 +182,40 @@ public static Builder builder() {
...
-    private final CascadingCompositeConfig     override;
+    private final CascadingCompositeConfig     dynamic;
     private final CascadingCompositeConfig     library;
     private final DefaultConfigLoader          loader;
     private final PropertyFactoryDynamicConfigListener dynamicObserver;
     
-    public DefaultAppConfig(Builder builder) {
-        super(NAME);
+    public DefaultAppConfig(Builder builder) throws ConfigException {
+        super("APP_CONFIG");
```
brharrington: There is still a `NAME` constant with this value, why is the string used directly?

archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
```diff
@@ -205,46 +182,40 @@ public static Builder builder() {
...
         
         try {
+            this.root = new CascadingCompositeConfig(NAME);
```
brharrington: Should this be using the `NAME` or `ROOT` constant?

archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
```diff
@@ -205,46 +182,40 @@ public static Builder builder() {
...
-            if (builder.enableSystemLayer) {
-                super.addConfig(new SystemConfig());
-            }
-            
-            if (builder.enableEnvironmentLayer) {
-                super.addConfig(new EnvironmentConfig());
-            }
+            root.addConfig(dynamic = new CascadingCompositeConfig(DYNAMIC_LAYER));
+            root.addConfig(new SystemConfig());
+            root.addConfig(new EnvironmentConfig());
```
brharrington: What is the rationale for not providing the option of leaving out system properties and environment variables? For test cases especially I would prefer to leave these out so I know my test case is isolated.

archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
```diff
@@ -253,8 +224,8 @@ public DefaultAppConfig(Builder builder) {
                     .withStrInterpolator(getStrInterpolator())
                     .build();
                 
-            super.addConfig(loader.newLoader().withName(APPLICATION_LAYER).load(builder.configName));
-            super.addConfig(library = new CascadingCompositeConfig(LIBRARY_LAYER));
+            root.addConfig(loader.newLoader().withName(APPLICATION_LAYER).load(builder.configName));
```
brharrington: I don't think this actually solves the problem Brent was asking about. He wants to set the Config used for the APPLICATION layer, or at least be able to add a Config to it if it was a composite.

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -171,8 +172,11 @@ final protected void configure() {
     
     @Provides
     @Singleton
-    final AppConfig getAppConfig() {
-        return DefaultAppConfig.createDefault();
+    final AppConfig getAppConfig(CascadeStrategy defaultStrategy) throws ConfigException {
+        return DefaultAppConfig.builder()
+                .withDefaultCascadingStrategy(defaultStrategy)
+                .withApplicationConfigName(DefaultAppConfig.DEFAULT_APP_CONFIG_NAME)
```
brharrington: Is there any reason to explicitly set to the default?

archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
elandau: yes

archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
elandau: Not sure how this happened.  Will fix.

archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
elandau: Was trying to simplify the knobs.  Will put this back.

archaius2-core/src/main/java/com/netflix/archaius/DefaultAppConfig.java
elandau: Agreed.  I'll make it so you can pass in your own Config for the application layer.

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
elandau: None.  Will remove.

elandau:
@brharrington, thanks for the thorough review.  I believe all your comments have been addressed but there is one thing I'm not too happy about.  Should we allow setting an external application layer Config AND try to load the configuration properties files as well.  Currently, when an external application config is provided DefaultAppConfig will silently not load the application properties.  Does it make sense to load both into a composite layer?

brharrington:
Can it just be a special case of calling `withApplicationConfig`? In which case if you do both it would be the same as calling `withApplicationConfig` twice on the same builder. It seems like this is mostly a question of what the builder makes convenient, if a user needs to have a config and load the configuration properties then they could just set the application layer to a composite themselves.

PR #301, opened 2015-05-01
Guice module to setup up Persisted2 client

elandau:
Persisted2ClientProvider and documentation to setup a Persisted2 client as the OverrideLayer

### DIFF
--- a/archaius2-persisted2/src/main/java/com/netflix/archaius/persisted2/Persisted2ConfigProvider.java
+++ b/archaius2-persisted2/src/main/java/com/netflix/archaius/persisted2/Persisted2ConfigProvider.java
@@ -68,7 +68,10 @@ public class Persisted2ConfigProvider implements Provider<Config> {
     public static String getFilterString(Map<String, List<String>> scopes) {
         StringBuilder sb = new StringBuilder();
         for (Entry<String, List<String>> scope : scopes.entrySet()) {
-            if (sb.length() > 0 && !scope.getValue().isEmpty()) {
+            if (scope.getValue().isEmpty()) 
+                continue;
+            
+            if (sb.length() > 0) {
                 sb.append(" and ");
             }
             
@@ -83,7 +86,7 @@ public class Persisted2ConfigProvider implements Provider<Config> {
                     first = false;
                 }
                 sb.append(scope.getKey());
-                if ("null".equals(value)) {
+                if (null == value) {
                     sb.append(" is null");
                 }
                 else if (value.isEmpty()) {
--- a/archaius2-persisted2/src/main/java/com/netflix/archaius/persisted2/Persisted2ConfigProvider.java
+++ b/archaius2-persisted2/src/main/java/com/netflix/archaius/persisted2/Persisted2ConfigProvider.java
@@ -1,0 +1,118 @@
+package com.netflix.archaius.persisted2;
+
+import java.net.URL;
+import java.net.URLEncoder;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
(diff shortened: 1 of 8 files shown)


archaius2-persisted2/src/main/java/com/netflix/archaius/persisted2/Persisted2ConfigProvider.java
```diff
@@ -0,0 +1,117 @@
...
+            .toString();
+        this.config = config;
+    }
+    
+    public static String getFilterString(Map<String, List<String>> scopes) {
+        StringBuilder sb = new StringBuilder();
+        for (Entry<String, List<String>> scope : scopes.entrySet()) {
+            if (sb.length() > 0 && !scope.getValue().isEmpty()) {
+                sb.append(" and ");
+            }
```
brharrington: If I'm reading this right it would add empty parens if one of the scopes is empty, e.g.:

```json
{
  "scope1": ["foo"],
  "scope2": [],
  "scope3": ["bar"]
}
```

Would lead to a query like:

```
(scope1='foo') () and (scope3='bar')
```

elandau: Good catch.  This really should just skip the scope if scope.getValue().isEmpty()

archaius2-persisted2/src/main/java/com/netflix/archaius/persisted2/Persisted2ConfigProvider.java
```diff
@@ -0,0 +1,117 @@
...
+            boolean first = true;
+            for (String value : scope.getValue()) {
+                if (!first) {
+                    sb.append(" or ");
+                }
+                else {
+                    first = false;
+                }
+                sb.append(scope.getKey());
+                if ("null".equals(value)) {
```
brharrington: Is there a reason for using the string null instead of having null mean null?

brharrington:
It would be nice if this provided more flexibility for the query so it could be configured to have the server filter out global properties. I currently generate a query that requires one of app, cluster, or asg to be set.

archaius2-persisted2/src/main/java/com/netflix/archaius/persisted2/Persisted2ConfigProvider.java
elandau: None.  I'll remove the string "null".

elandau:
Persisted2ConfigProvider doesn't force any particular query scope structure so you could set the query scope to anything you like. 

Here's how you can set up a DefaultPersisted2ClientConfig for just 'appId', 'cluster' and 'asg' and no global scope.

```java
final Persisted2ClientConfig config = new DefaultPersisted2ClientConfig()
            .withQueryScope("appId",    "app1")
            .withQueryScope("cluster", "cluster1")
            .withQueryScope("asg", "asg1");
```

This will translate to the following filter,
filter=(app='app1') and (cluster='cluster1') and (asg='asg1')

brharrington:
Right, but that will only match properties with all three fields explicitly set on the server. If I set a property with just the application specified, then it won't match. We currently generate a query that looks like:

```
(
  (asg is null or asg='' or asg='asg1') or
  ((asg is null or asg='') and (cluster is null or cluster='' or cluster='cluster1')) or
  ((asg is null or asg='') and (cluster is null or cluster='') and app='app1')
) and ...
```

Without something like this a user would have to be very careful with how it was entered into platformservice or result to using client side filtering which means thousands of global properties that will get ignored are pulled down on every update.

PR #303, opened 2015-05-05
First pass at sub project to bridge 1.x and 2.x

elandau:
Bridging enabled by installing the StaticArchaiusBridgeModule in addition to ArchaiusModule

```java
Guice.createInjector(new ArchaiusModule(), new StaticArchaiusBridgeModule());
```


archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/AbstractConfigurationBridge.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Override
+    public void addConfiguration(AbstractConfiguration config) {
+        addConfiguration(config, "Config-" + libNameCounter.incrementAndGet());
+    }
+
+    @Override
+    public void addConfiguration(AbstractConfiguration config, String name) {
+        try {
+            libraries.addConfig(name, new CommonsToConfig(config));
```
brharrington: When using archaius1 if I call:

```java
ConfigurationManager.loadPropertiesFromResources("a.properties");
ConfigurationManager.loadPropertiesFromResources("b.properties");
```

The settings from `b.properties` will override settings from `a.properties`. This implementation reverses that. In some cases we rely on that ordering for things like account specific overrides.

elandau: Are you sure about this behavior.  I just wrote a test that confirms the opposite.

brharrington: I tried to reproduce in a simple test case using the OSS lib directly and A overrides B. However, if I use NFLifecycleUnitTester so it gets bound to NetflixConfiguration I see B override A.

elandau: Oh, great!!!  OK.  I'm gonna make the override order in a CompositeConfig configurable.

brharrington: I'm not sure that will help. My goal is to have consistent behavior and make sure the libraries are doing the right thing (or at least what they are expected to do) when using archaius1, archaius1 w/ netflix-configuration, or archaius1 over archaius2. From the perspective of a user the order cannot be relied on because I don't necessarily control how the configuration was setup. 

I'll update my use-cases so I don't make any assumptions about the order. Not sure how much this will impact others, but I'm guessing there are quite a few places where changing the order will break existing apps either intentionally or accidentally. If possible for the bridge it seems like matching the behavior of NetflixConfiguration would probably cause the least surprise.

elandau: Specifying the order will be a setting for the @LibrariesLayer CompositeConfig that is done in the archaius-guice integration and will therefore be separate from the bridge.  It will be an application decision.  So, for internal Netflix we would set this up with reversed order to match the behavior of NetflixConfiguration.  This will be the case both for the bridge as well as regular use of archaius2.

brharrington: Personally, I would prefer archaius2 have a single order that is always used for a CompositeConfig as it is easier to reason about. Is the reverse order change actually desirable for any reason other than trying to mitigate the pain of migrating from archaius1?

elandau: Honestly, I'd rather not even have the libraries layer and keep all libraries fully isolated.  However, unless we start supporting remote overrides with library scope this would make overrides cumbersome and difficult to reason.  Unfortunately we do have to deal with the reality that there are two different override orders in use today.  The default will remain consistent with the previous OSS but I do like at least having the option to control this feature.

brharrington: > Unfortunately we do have to deal with the reality that there are two different override orders in use today.

With the legacy library that is true, but it could be dealt with specific to the bridge rather than making it true when using 2.x directly. To me this seems like a confusing aspect of archaius1. 

In terms of archaius2, my concern is that as a library author I can't make a lot of assumptions about how an application owner will setup the Config. Before if I had code like:

```java
@Inject
public Foo(@LibraryLayer CompositeConfig config) {
  config.addConfig(getConfig("a.properties"));
  config.addConfig(getConfig("b.properties"));
}
```

I could tell that A would override B. Now a library owner can't determine which one will override. I can work around the problem by creating my own composite instance and adding that as a single addition to the library layer. It just seems like this adds a lot of potential for mistakes and confusion when using archaius2.

elandau: Regardless of override order there can still be a lot of ambiguity stemming from the order in which classes are instantiated.  Take the following two classes for example,

```java
@ConfigurationSource("a")
public class AConfiguration {
}

@ConfigurationSource("b")
public class BConfiguration {
}
```

Since AConfiguration and BConfiguration have no dependencies between them they can be instantiated in any random order so the override behavior is essentially undefined.  When you have more control over loading the property files it might actually make perfect sense to load into a child CompositeConfig (as you described) where you have full control over the override order.  

Perhaps we need to come up with a configuration dependency mechanism that could potentially re-order child configurations after they are loaded.  I'm not sure what that would look like though and am open to suggestions.

Issue #305, opened 2015-05-14
2.x what is the recommended way to use the archaius1 bridge ?


brharrington:
I've been trying to use the archaius1 bridge, but it doesn't work if the static block of ConfigurationManager gets executed before the configure method of StaticArchaiusBridgeModule. That is pretty fragile in practice and typically it means archaius1 is getting mapped to the default implementation. 

If I set the system properties being set by the module on the java command I get an NPE because the static injection hasn't occurred yet:

```
Caused by: java.lang.RuntimeException: Error initializing configuration
	at com.netflix.config.ConfigurationManager.<clinit>(ConfigurationManager.java:109)
	... 28 more
Caused by: java.lang.NullPointerException
	at com.netflix.archaius.bridge.StaticDeploymentContext.getValue(StaticDeploymentContext.java:71)
	at com.netflix.config.ConfigurationManager.setDeploymentContext(ConfigurationManager.java:351)
	at com.netflix.config.ConfigurationManager.<clinit>(ConfigurationManager.java:96)
	... 28 more
```

elandau:
I noticed this as well.  I need to modify the 0.x code to allow this.  Currently, once the static block has been called it's impossible to change.

elandau:
I could modify the bridge to not set the archaius 0.x system properties but instead call,
```java
ConfigurationManager.install(this);
```
from the constructor of the bridge.

elandau:
Fixed in #306

PR #306, opened 2015-05-14
Fix for #305

elandau:
The bridge will now configure the old ConfigurationManager API using install() instead of changing the default implementation via properties, "archaius.default.configuration.class" and "archaius.default.deploymentContext.class".  This way the default AbstractConfiguration implementation may be replaced once the injector is created without throwing exceptions.  Any code accessing ConfigurationManager prior to the injector created will use the default AbstractConfiguration implementation and DeploymentContext.  It is still strongly encouraged to NOT access ConfigurationManager prior to the injector being created.


brharrington:
I pulled this into my test app and it seems to be working. It also exposed some issues for properties with `@` and typesafe config (#307).

PR #309, opened 2015-05-21
Make threads daemon.  Expose method to shutdown pollers.

elandau:


archaius2-core/src/main/java/com/netflix/archaius/config/polling/FixedPollingStrategy.java
```diff
@@ -39,8 +40,9 @@ public FixedPollingStrategy(long interval, TimeUnit units) {
     }
     
     public FixedPollingStrategy(long interval, TimeUnit units, boolean syncInit) {
+        this.executor = Executors.newSingleThreadScheduledExecutor(ThreadFactories.newNamedDaemonThreadFactory("poller-%d"));
```
brharrington: It would be handy to have archaius in the thread name so it is easier to spot at a glance and realize it is an archaius poller and not a poller for something else.

PR #318, opened 2015-06-12
EtcdConfigurationSource

spoon16:
An Etcd watched configuration source very similar to the ZooKeeperConfigurationSource.

Uses the Boon Etcd client https://github.com/boonproject/boon/blob/master/etcd/README.md.  Seems to be the most widely adopted Etcd client.

Tests included.

### DIFF
--- a/archaius-etcd/src/main/java/com/netflix/config/source/EtcdConfigurationSource.java
+++ b/archaius-etcd/src/main/java/com/netflix/config/source/EtcdConfigurationSource.java
@@ -1,0 +1,147 @@
+package com.netflix.config.source;
+
+import com.google.common.base.Objects;
+import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Maps;
+import com.netflix.config.WatchedConfigurationSource;
+import com.netflix.config.WatchedUpdateListener;
+import com.netflix.config.WatchedUpdateResult;
+import org.boon.core.Handler;
+import org.boon.etcd.Etcd;
+import org.boon.etcd.Node;
+import org.boon.etcd.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+
+/**
+ * Implementation of the dynamic {@link WatchedConfigurationSource} for Etcd
+ *
+ * This implementation requires the path to the Etcd directory that contains
+ * nodes that represent each managed configuration property.
+ * 
+ * An example Etcd configuration path is /<my-app>/config
+ * An example Etcd property node path is /<my-app>/config/com.fluxcapacitor.my.property
+ * 
+ * When a property is mutated via Etcd a callback will be notified and the value managed
+ * by EtcdConfigurationSource will be updated. Similar to other dynamic configuration
+ * source (ie. DynamoDB, etc.)
+ * 
+ * @author spoon16
+ */
+public class EtcdConfigurationSource implements WatchedConfigurationSource {
+    private static final Logger logger = LoggerFactory.getLogger(EtcdConfigurationSource.class);
+    private static final Splitter keySplitter = Splitter.on('/');
+
+    private final Map<String, Object> valueCache = Maps.newConcurrentMap();
+    private final List<WatchedUpdateListener> listeners = new CopyOnWriteArrayList<WatchedUpdateListener>();
+
+    private final Etcd etcd;
+    private final String configPath;
+
+    private Handler<Response> updateHandler = new Handler<Response>() {
+        @Override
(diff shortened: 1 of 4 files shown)


spoon16:
Build failed because of a JVM error... or maybe a bunch of failing unit tests unrelated to the Etcd module.

Unless I am missing something.

howardyuan:
There're two pull-requests builds. One of them was not configured with
proper Gradle opts so it runs out Permgen. I'll see how to either disable
the second one or add proper GRADLE_OPTS.

archaius-etcd/src/main/java/com/netflix/config/source/EtcdConfigurationSource.java
```diff
@@ -0,0 +1,147 @@
...
+        public void handle(Response updateResponse) {
+            if (updateResponse.wasError()) {
+                logger.error("Etcd failed with an error response: %s", updateResponse);
+            }
+
+            final Map<String, Object> create = Maps.newHashMap();
+            final Map<String, Object> set = Maps.newHashMap();
+            final Map<String, Object> delete = Maps.newHashMap();
+
+            final String action = updateResponse.action().toLowerCase();
```
howardyuan: Can the action() return null when it's error?

archaius-etcd/src/main/java/com/netflix/config/source/EtcdConfigurationSource.java
```diff
@@ -0,0 +1,147 @@
...
+    private final String configPath;
+
+    private Handler<Response> updateHandler = new Handler<Response>() {
+        @Override
+        public void handle(Response updateResponse) {
+            if (updateResponse.wasError()) {
+                logger.error("Etcd failed with an error response: %s", updateResponse);
+            }
+
+            final Map<String, Object> create = Maps.newHashMap();
```
howardyuan: I was a little puzzled by three maps here for a while. For sure, we will get two empty maps and one map with one entry in there? Not really sure if I have a better idea yet, just want to point out.

archaius-etcd/src/main/java/com/netflix/config/source/EtcdConfigurationSource.java
```diff
@@ -0,0 +1,147 @@
...
+            final Map<String, Object> delete = Maps.newHashMap();
+
+            final String action = updateResponse.action().toLowerCase();
+            final Node node = updateResponse.node();
+
+            if (node != null ) {
+                final String etcdKey = node.key();
+                final String sourceKey = Iterables.getLast(keySplitter.split(etcdKey));
+                final String value = node.getValue();
+                valueCache.put(sourceKey, value);
```
howardyuan: what if it's a delete, then should you remove the entry from the cache?

spoon16: good catch, will fix

archaius-etcd/src/main/java/com/netflix/config/source/EtcdConfigurationSource.java
spoon16: Yeah, not sure what would be better here. I could do something like:

```java

```

archaius-etcd/src/main/java/com/netflix/config/source/EtcdConfigurationSource.java
spoon16: only if I duplicate the etcd.waitRecursive call on l:89

It'd probably be best to capture the wasError condition, throw an exception and have a finally that calls etcd.waitRecursive.  But I'm not sure that really makes things more readable.  I'll just duplicate l:89 and return early.

archaius-etcd/src/main/java/com/netflix/config/source/EtcdConfigurationSource.java
spoon16: fixed and added a test to validate

spoon16:
@howardyuan, tried to address your feedback with the following changes.

- fixed bug where delete wouldn't remove the value from the `EtcdConfigurationSource`
- added a unit test to validate delete behavior
- refactored the `updateHandler` in `EtcdConfigurationSource`... tried to improve readability by not always instantiating three maps

howardyuan:
Looks good to me. Will leave the PR open for a few days to see if anybody else have any more comments.

chandrapersonal:
can i pull the changes.

spoon16:
@howardyuan merge?

chandrapersonal:
why there is no case for node action "compareandswap" in etcd UpdateHandler. we have node actions create,set and delete action on UpdateHandler . but no  compareandswap. Is there any reason for not have that action.

spoon16:
@chandrapersonal I don't think "compareandswap" is a type of update.  It's part of the Etcd update API and the side effect is that an update notification is emitted, either set or create.

PR #323, opened 2015-06-24
Enhancements to configuration proxy

elandau:
* Add PropertyInfo annotation to override the property name when proxying
* Add support for making proxied configuration immutable
* Add toString to proxies configuration interfaces

Example of new features,
```java
@Configuration(immtuable=true) // Once created properties never change
public interface MyConfig {
   // Looks for property named "foo.bar" instead of "fooBar" (derived from name)
   @PropertyInfo("foo.bar")
   public String getFooBar();

   // 'is' verb supported in addition to 'get'
   public Boolean isEnabled();
}
```

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/annotations/PropertyInfo.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/annotations/PropertyInfo.java
@@ -23,6 +23,6 @@ import java.lang.annotation.Target;
 @Documented
 @Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
 @Target({ElementType.FIELD, ElementType.METHOD})
-public @interface Value {
+public @interface PropertyInfo {
     String name();
 }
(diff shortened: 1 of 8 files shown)


archaius2-core/src/main/java/com/netflix/archaius/annotations/PropertyInfo.java
```diff
@@ -23,6 +23,6 @@
 @Documented
 @Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
 @Target({ElementType.FIELD, ElementType.METHOD})
-public @interface Value {
+public @interface PropertyInfo {
```
qiangdavidliu: Would PropertyName be more intuitive?

elandau: Ya, I don't like PropertyInfo.  However, I envision adding other attributes to this annotation.  Alternatively we can make the decision that PropertyName will only be used for naming and as such cannot have other attributes.  Additional attributes like immutability can be specified using a different annotation such as @Immutable.

qiangdavidliu: @ Property is not yet taken. If you want a more uber annotation that allows definition of alternative names and optionally other attributes, how about that?

PR #324, opened 2015-06-28
Cleanup and simplify usage of ConfigProxyFactory

elandau:
The previous interface and method of constructing a configuration proxy wasn't very clean and could result in excessive property objects and listeners being created.  The new approach attaches the ConfigProxyFactory (previously ProxyFactory) directly to the top level PropertyContainerFactory.

Creating a proxy in Guice can now be done in one line.

Example,
```java
@Provides
@Singleton
MyConfig getMyConfig(ConfigProxyFactory factory) {
    return factory.newProxy(MyConfig.class);
}
```

In addition, a prefix can be specified for all proxy properties.

```java
@Provides
@Singleton
MyConfig getMyConfig(ConfigProxyFactory factory) {
    return factory.newProxy(MyConfig.class, "someprefix");
}
```


tbak:
This change makes proxy creation much simpler and compact.

Does it make sense to have an interface for ConfigProxyFactory? Current implementation is tied to Java-style properties. For JSON configuration files or INI  file format one might wish to provide different mapping mechanism. ```PropertyFactory``` has more generic contract, which does not enforce a specific property name formatting.

There is a TODO about subproxy. It would be very useful feature, that would allow for hierarchical configuration. Any chance to get it soon? For example Eureka2 has a configuration that is actually a composition of a few different, subsystem specific configurations, but now merged into single class implementing all configuration interfaces.

elandau:
Previously ConfigProxyFactory was an interface but I didn't see much value in that since there would only be one implementation.  The abstraction for different configuration formats (JSON, INI, etc.) is already handled int he Config interface since all configuration formats can ultimately be reduced to a tree structure where a traditional Java property maps to a path through the tree. 

I'll work on subproxy next.

PR #326, opened 2015-06-29
Inheritance support in configuration proxying

elandau:
* Support proxying methods from inherited interfaces
* Remove incorrect setting of default value when parsing non-primitive types in fast properties
* Delegate parsing of various types to Decoder

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/ProxyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/ProxyFactory.java
@@ -3,13 +3,10 @@ package com.netflix.archaius;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
-import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


archaius2-core/src/main/java/com/netflix/archaius/ProxyFactory.java
```diff
@@ -68,6 +70,24 @@ private String derivePrefix(Configuration annot, String prefix) {
         return prefix + ".";
     }
     
+    /**
+     * Populate a set of all interface and sub interface methods via recursion.
+     * Give top level overrides priority over methods defined in super-interfaces
+     * @param type
+     * @param methods
+     */
+    private void fillConfigMethods(Class<?> type, Set<Method> methods) {
```
tbak: You can get all methods via Class.getMethods() call.

elandau: True.  I was following a pattern from governator.  I need to understand why governator didn't use getMethods().

PR #327, opened 2015-07-07
Read properties resource with utf-8 encoding.

killme2008:
We have some property values in chinese, but it seems that archaius read the properties with default IOS-8859-1 encoding, then the values in chinese become unreadable code. 
We have to cast the chinese values to escaped unicode,it's really not human-friendly.

I make some changes in this PR to fix this problem, that it use `InputStreamReader` with utf-8 encoding to load `Properties`, and it resolve our problem.


archaius-core/src/main/java/com/netflix/config/ConfigurationManager.java
```diff
@@ -241,8 +242,17 @@ public static void loadPropertiesFromResources(String path)
...
-            props.load(fin);
-            fin.close();
+            InputStreamReader reader = new InputStreamReader(fin, "UTF-8");
+            try {
+              props.load(reader);
+            } finally {
+              if (reader != null) {
+                reader.close();
+              }
+              if (fin != null) {
```
elandau: fin can never be null

archaius-core/src/main/java/com/netflix/config/ConfigurationManager.java
```diff
@@ -241,8 +242,17 @@ public static void loadPropertiesFromResources(String path)
             URL url = resources.nextElement();
             Properties props = new Properties();
             InputStream fin = url.openStream();
-            props.load(fin);
-            fin.close();
+            InputStreamReader reader = new InputStreamReader(fin, "UTF-8");
```
howardyuan: This code path is seldom used. But I'm fine with fixing it here as well. Please consolidate the functionality here and put it in a function inside ConfigurationUtils class so we don't have duplicate code in two places.

archaius-core/src/main/java/com/netflix/config/sources/URLConfigurationSource.java
```diff
@@ -180,8 +181,17 @@ public PollResult poll(boolean initial, Object checkPoint)
         for (URL url: configUrls) {
             Properties props = new Properties();
             InputStream fin = url.openStream();
-            props.load(fin);
-            fin.close();
+            InputStreamReader reader = new InputStreamReader(fin, "UTF-8");
```
howardyuan: This is a biggie. I didn't notice this before because we didn't run into any CJK values before. In current code path, the most frequently used file properties loading is through PropertiesConfiguration in Apache Commons Configurations. But I like not depending on that and doing it this way. The only comment I have here is as above. Consolidate the code into a function in ConfigurationUtils so we can re-use.

killme2008: Thanks for your reviews, i will fix them and update the PR.

killme2008:
@howardyuan @elandau   Updated the PR, please review it again.

gorzell:
Looks like this will need to be patched into the 2.x branch as well.

PR #328, opened 2015-07-08
First draft of making MapConfig generic.

gorzell:
In the current implementation MapConfig is forced to be String, String.  I don't think this makes sense for all cases and have a specific case where it would be nice if it was String, Object.  It seemed like it might make more sense to add a type parameter than proliferate MapConfig classes of various types.  I tried to leave as much of the functionality intact as possible.

If the String flavor is highly used it would be an option to make a child class that uses that type.

There is still one test broken, but I wanted to get a draft up for discussion.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/config/MapConfig.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/config/MapConfig.java
@@ -25,7 +25,7 @@ import java.util.Properties;
 /**
  * Config backed by an immutable map.
  */
-public class MapConfig extends AbstractConfig {
+public class MapConfig<T> extends AbstractConfig {
 
     /**
      * The builder only provides convenience for fluent style adding of properties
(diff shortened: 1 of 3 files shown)


elandau:
One big issue with this is that all value extraction mechanisms assume the value is a String and not some other type (ex. Integer, List etc...).  We would have to modify all parsing code to determine the type of data being stored.  Support for value types other than String and especially collections is a key missing feature from archaius2 and now might be a good time to address that.  Can you add unit tests for storing a value type other than String (ex. Integer) as well as storing a List.  That should at least help illustrate that more work needs to be done.

gorzell:
That would def be a problem.  I guess I assumed based on the API that it was the other way around, but I can also see how that would be hard.

My specific case is trying to deal with a YAML config file.  In the short term I was going to hand it by loading into a map, but in the long term it would probably warrant a full plugin.

I am happy to add the unit tests.

elandau:
You can load a YAML config file using the Typesafe Config integration.  I do however know that archaius2 has issues with dealing with collections.

gorzell:
I'll take a look at that, it may or may not work because we have multiple maps (one per env) with a configuration in each.

gorzell:
I am also not terribly opposed to a decoding behavior where if you asked for something in a type that it isn't you log an error and return the default or throw an exception.

PR #337, opened 2015-07-29
Nullable and argument replacement for configuration proxy

elandau:
Add support for @Nullable property values.
```java
@Nullable
public getSomeProperty();
```

Add support to use arguments passed to interface getters as replacements for deriving the property names.
```java
@PropertyName(name="foo.${0}.${1}")
public getFoo(String arg0, String arg1);
```

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/ProxyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/ProxyFactory.java
@@ -8,6 +8,7 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
 
+import javax.annotation.Nullable;
 import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


archaius2-core/src/main/java/com/netflix/archaius/ProxyFactory.java
```diff
@@ -11,9 +11,12 @@
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import org.apache.commons.lang3.text.StrSubstitutor;
+
 import com.netflix.archaius.annotations.Configuration;
 import com.netflix.archaius.annotations.DefaultValue;
 import com.netflix.archaius.annotations.PropertyName;
+import com.sun.istack.internal.Nullable;
```
spencergibb: Is that the right `Nullable`?

PR #345, opened 2015-09-11
Fix Configuration loading order; add support for list of @next and in…

howardyuan:
…terpolation for @next resources


archaius2-core/src/main/java/com/netflix/archaius/readers/PropertiesConfigReader.java
```diff
@@ -32,38 +35,46 @@
...
         if (url == null) {
             throw new ConfigException("Unable to resolve URL for resource " + resourceName);
         }
-        return load(loader, url);
+        return load(loader, url, strInterpolator, lookup);
     }
 
     @Override
-    public Config load(ClassLoader loader, URL url) throws ConfigException {
+    public Config load(ClassLoader loader, URL url, StrInterpolator strInterpolator, StrInterpolator.Lookup lookup) throws ConfigException {
```
elandau: There is a potential bug here when having files nested with @next commands where @next in a nested file will go to the end of the list.  It might be better to use recursion here.  We should also check for circular @next dependencies as well.

howardyuan: K, just submitted a fix to switch to depth first instead of width first loading. Also adding the check for the infinite chaining.

archaius2-core/src/main/java/com/netflix/archaius/readers/PropertiesConfigReader.java
```diff
@@ -32,38 +32,57 @@
...
-            
-            return new MapConfig(props);
-        }
-        catch (IOException e) {
-            throw new ConfigException("Unable to load configuration " + url, e);
+                if (next != null) {
+                    String[] listOfAtNext = next.split(",");
+                    LinkedList<String> listToAdd = new LinkedList<>();
+                    for (String urlString : listOfAtNext) {
+                        String extrapolatedForm = strInterpolator.create(lookup).resolve(urlString);
```
elandau: s/extrapolated/interpolated

archaius2-core/src/main/java/com/netflix/archaius/readers/PropertiesConfigReader.java
```diff
@@ -32,38 +32,57 @@
...
+                        if (!existingUrls.contains(extrapolatedForm)) {
+                            existingUrls.add(extrapolatedForm);
+                            listToAdd.add(extrapolatedForm);
```
elandau: Why not just add directly to nextUrls.

PR #346, opened 2015-09-11
Minor cleanup to archaius1 bridge

elandau:


archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/StaticAbstractConfiguration.java
```diff
@@ -27,11 +26,7 @@
 
     @Inject
     public static void intialize(AbstractConfigurationBridge config) {
-        LOG.info("Initializing");
         delegate = config;
-        if (!ConfigurationManager.isConfigurationInstalled()) {
-            ConfigurationManager.install(config);
```
brharrington: Does this change the way the bridge module is supposed to be used? In one of our apps it looks like the bridge isn't working after picking up a more recent RC.

PR #347, opened 2015-09-23
More robust StaticAbstractConfiguration bridge

elandau:
Remove the concept of application override layer from ArchaiusConfiguration.  Application overrides should be dealt with in the ApplicationLayer

### DIFF
--- a/archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/StaticAbstractConfiguration.java
+++ b/archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/StaticAbstractConfiguration.java
@@ -1,17 +1,19 @@
 package com.netflix.archaius.bridge;
 
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 import javax.inject.Inject;
 
 import org.apache.commons.configuration.AbstractConfiguration;
 import org.apache.commons.configuration.Configuration;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.netflix.config.AggregatedConfiguration;
+import com.netflix.config.ConfigurationManager;
+import com.netflix.config.DynamicPropertyFactory;
 import com.netflix.config.DynamicPropertySupport;
 import com.netflix.config.PropertyListener;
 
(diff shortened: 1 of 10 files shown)


archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/StaticAbstractConfiguration.java
```diff
@@ -20,13 +22,35 @@
...
+    public StaticAbstractConfiguration() {
+        staticConfig = this;
+    }
+    
     @Inject
     public static void intialize(AbstractConfigurationBridge config) {
         delegate = config;
+
+        if (staticConfig == null) {
+            System.err.println("Trying to use bridge but but hasn't been configured yet!!!");
```
brharrington: s/but but/but it/

archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/StaticAbstractConfiguration.java
```diff
@@ -35,21 +59,37 @@ public static void reset() {
 
     @Override
     public boolean isEmpty() {
+        if (delegate == null) {
+            System.err.println("[isEmpty()] StaticAbstractConfiguration not initialized yet.");
```
brharrington: If the config is broken would it make more sense to throw an exception? The stderr maybe directed somewhere other than the primary log and so it seems like this would be easy to miss. And if the application isn't configured correctly failing outright would probably be better than coming up in an odd state with strange or broken behavior because configuration wasn't working correctly.

archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/StaticDeploymentContext.java
```diff
@@ -89,7 +89,7 @@ public void setDeploymentRegion(String region) {
     @Override
     public String getValue(ContextKey key) {
         if (delegate == null) {
-            LOG.warn("Configuration not yet initialized.  Returning 'null' for " + key);
+            LOG.debug("Configuration not yet initialized.  Returning 'null' for " + key);
```
brharrington: Just curious, why does this class use logging but StaticAbstractConfiguration is being changed to System.err?

archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/StaticAbstractConfiguration.java
elandau: Due to the excessive use of statics in various libraries (blitz4j) there are some unavoidable situations where the configuration system hasn't been set up yet and it would be impossible to move forward if this code threw an exception.  For now I'm opting to output an error and slowly fix libraries that have these issues.  In most cases the libraries have very reasonable defaults or get reconfigured later on.

PR #348, opened 2015-09-24
Improvements to .properties file loading

elandau:
* Change property file loading behavior to succeed whether the first file exists or not.  Deprecate ConfigLoader.Loader.withFailOnFirst
* Simplify @next support implementation using recursion
* '.properties' is optional for files listed in @next


howardyuan:
LGTM

Issue #350, opened 2015-10-02
Archaius2 config.getKeys(String prefix) returns an infinite iterator


howardyuan:
config.getKeys(String prefix) is supposed to return the iterator for the properties start with the prefix. But there's a bug that the results are in an infinite loop.

PR #357, opened 2015-10-12
issue #350. Fix the bug in key prefix iterator in Config

howardyuan:
This is actually pretty tricky. There are a couple of issues prior to this. One is at the beginning, the other is at the end of underneath iterator. We need to proceed until we can find one matching prefix at the beginning. At the end, we need to terminate properly by setting the "next" to null at proper place. It's very easy to miss the final one element.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
@@ -162,8 +162,7 @@ public abstract class AbstractConfig implements Config {
         return new Iterator<String>() {
             Iterator<String> iter = getKeys();
             String next;
-            String prev;
-            
+
             {
                 while (iter.hasNext()) {
                     next = iter.next();
@@ -187,7 +186,8 @@ public abstract class AbstractConfig implements Config {
                     throw new IllegalStateException();
                 }
                 
-                prev = next;
+                String current = next;
+                next = null;
                 while (iter.hasNext()) {
                     next = iter.next();
                     if (next.startsWith(prefix)) {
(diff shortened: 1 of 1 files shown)


elandau:
Is there a unit test to show the behavior and confirm the fix?

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -162,10 +162,17 @@ public String getString(String key) {
         return new Iterator<String>() {
             Iterator<String> iter = getKeys();
             String next;
+            String prev;
```
elandau: No need for this to be a field

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -190,6 +197,9 @@ public String next() {
                         next = null;
                     }
                 }
+                if (next != null && next.equals(prev)) {
```
elandau: We can eliminate this extra check, 

```java
String current = next;
next = null;
while (iter.hasNext()) {
    next = iter.next();
    if (next.startsWith(prefix)) {
        break;
    }
    else {
         next = null;
    }
}
return current;
````

**Comment by elandau** [comment]
Good catch on the startup case!

File: archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java [hunk 3]
```diff
@@ -162,10 +162,16 @@ public String getString(String key) {
         return new Iterator<String>() {
             Iterator<String> iter = getKeys();
             String next;
-            
+
             {
-                if (iter.hasNext()) {
+                while (iter.hasNext()) {
```
elandau: This block of code repeats in next() so let's move it to a function.

Issue #354, opened 2015-10-02
Archaius2 Add ability to set/disable list delimiter for a property with delimiters


howardyuan:
Currently, Archaius2 hardcode comma as delimiter to delimit a property into an array. Should ability to Set different delimiters or disable delimiting all together.

howardyuan:
The set delimiter should be limited to the scoped to each config implementation instead of Composite Config level.

howardyuan:
PR has been merged.

Issue #353, opened 2015-10-02
Archaius2 Support arbitrary type in SetProperty. 


howardyuan:
Currently, Archaius2 SetProperty converts everything into String. We need to support the arbitrary object as the property type in SetProperty without converting it.

howardyuan:
PR has been merged.

PR #360, opened 2015-10-14
issues #353, #354. Adding ability to set new list delimiter and prope…

howardyuan:
…rty is stored as Object instead of String.

This PR covers two issues. One is to be able to set a new delimiter. The other is treat Property value as Object instead of String. For now, only turns on the storing for Settable config. In the future, it's possible we need to extend to MapConfig.

### DIFF
--- a/archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
+++ b/archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
@@ -69,7 +69,14 @@ public class CommonsToConfig extends AbstractConfig {
 ;
         List<T> result = new ArrayList<T>();
         for (Object part : value) {
-            result.add(getDecoder().decode(type, part.toString()));
+            if (type.isInstance(part)) {
+                result.add((T)part);
+            } else if (part instanceof String) {
+                result.add(getDecoder().decode(type, (String) part));
+            } else {
+                throw new UnsupportedOperationException(
+                        "Property values other than " + type.getCanonicalName() +" or String not supported");
+            }
         }
         return result;
     }
--- a/archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
+++ b/archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
@@ -15,7 +15,10 @@
  */
 package com.netflix.archaius.commons;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
+import java.util.List;
 
 import org.apache.commons.configuration.AbstractConfiguration;
 
@@ -55,4 +58,64 @@ public class CommonsToConfig extends AbstractConfig {
     public Iterator<String> getKeys() {
         return config.getKeys();
     }
+
+    @Override
+    public <T> List<T> getList(String key, Class<T> type) {
+        List value = config.getList(key);
+        if (value == null) {
+            return notFound(key);
+        }
+;
+        List<T> result = new ArrayList<T>();
+        for (Object part : value) {
+            result.add(getDecoder().decode(type, part.toString()));
+        }
+        return result;
+    }
+
+    @Override
+    public List getList(String key) {
+        List value = config.getList(key);
+        if (value == null) {
+            return notFound(key);
+        }
+        return value;
+    }
+
+    @Override
+    public String getString(String key, String defaultValue) {
+        List value = config.getList(key);
+        if (value == null) {
+            return notFound(key, defaultValue != null ? getStrInterpolator().create(getLookup()).resolve(defaultValue) : null);
+        }
+        List<String> interpolatedResult = new ArrayList<>();
+        for (Object part : value) {
+            if (part instanceof String) {
+                interpolatedResult.add(getStrInterpolator().create(getLookup()).resolve(part.toString()));
+            } else {
+                throw new UnsupportedOperationException(
+                        "Property values other than String not supported");
+            }
+        }
+        return String.join(getListDelimiter(), interpolatedResult);
+    }
+
+    @Override
+    public String getString(String key) {
+        List value = config.getList(key);
+        if (value == null) {
+            return notFound(key);
+        }
+        List<String> interpolatedResult = new ArrayList<>();
+        for (Object part : value) {
+            if (part instanceof String) {
+                interpolatedResult.add(getStrInterpolator().create(getLookup()).resolve(part.toString()));
+            } else {
+                throw new UnsupportedOperationException(
+                        "Property values other than String not supported");
+            }
+        }
+        return String.join(getListDelimiter(), interpolatedResult);
+    }
+
 }
--- a/archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
+++ b/archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
@@ -23,6 +23,7 @@ import java.util.List;
 import org.apache.commons.configuration.AbstractConfiguration;
 
 import com.netflix.archaius.config.AbstractConfig;
+import org.apache.commons.lang.StringUtils;
 
 /**
  * Adaptor to allow an Apache Commons Configuration AbstractConfig to be used
@@ -97,7 +98,7 @@ public class CommonsToConfig extends AbstractConfig {
                         "Property values other than String not supported");
             }
         }
-        return String.join(getListDelimiter(), interpolatedResult);
+        return StringUtils.join(interpolatedResult, getListDelimiter());
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -225,243 +227,158 @@ public Config getPrefixedView(String prefix) {
...
             return notFound(key);
-        try {
-            return decoder.decode(Long.class, value);
         }
-        catch (NumberFormatException e) {
-            return parseError(key, value, e);
+        if (rawProp instanceof String) {
+            try {
+                String value = interpolator.create(lookup).resolve(rawProp.toString());
+                if (type.equals(boolean.class) || type.equals(Boolean.class)) {
```
elandau: This should be part of the decoder

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -225,243 +227,158 @@ public Config getPrefixedView(String prefix) {
...
     public Long getLong(String key, Long defaultValue) {
-        String value = getString(key, null);
-        if (value == null) 
-            return defaultValue;
         try {
-            return decoder.decode(Long.class, value);
-        }
-        catch (NumberFormatException e) {
-            return parseError(key, value, e);
+            return getValue(Long.class, key);
```
elandau: Can we create a variant of getValue(Class<T> type, String key, T defaultValue) that will handle the NoSuchElementException instead of having to handle the exception in all the getXXX methods?

brharrington:
Just curious why does archaius core have a concept of list delimiter and parsing at all? Since how a list is structured and parsed would depend on the file format, it seems like that should be left to the config implementations such as archaius2-commons-configuration, archaius2-typesafe, archaius2-json, archaius2-yaml, etc.

archaius2-core/src/main/java/com/netflix/archaius/config/CompositeConfig.java
```diff
@@ -230,6 +230,14 @@ public Config getConfig(String name) {
     }
 
     @Override
+    public void setListDelimiter(String newDelimiter) {
```
elandau: Also add setListDelimiter to postConfigAdded() which is called for any child that is added.  This will address the case where a child is added after setListDelimiter() is called

howardyuan:
@brharrington This is for the feature parity of Archaius1 and Archaius2. Archaius 1 inherited from Apache Commons the behavior of treating the string property as list internally. So the list delimiting comes with the core package in Archaius1.

brharrington:
As a user I would expect the notion of a list to be based on the implementation used. So if I bind it to Apache commons I can set the delimiter using the commons configuration api, e.g.: 

```java
    public static void main(String[] args) {
        Map<String, Object> data = new HashMap<>();
        data.put("foo", "a,b");
        AbstractConfiguration commonsCfg = new MapConfiguration(data);
        commonsCfg.setListDelimiter(',');
        System.out.println("  commons:" + commonsCfg.getList("foo"));

        Config config = new CommonsToConfig(commonsCfg);
        System.out.println("archaius2:" +config.getList("foo"));
    }
```

Then when using the archaius2 api getList would be equivalent to the corresponding call of the underlying implementation. And at least right now that doesn't seem to work: 

```
  commons:[a, b]
archaius2:[a]
```

In general `setListDelimiter` seems odd to me on the core `Config` interface. To implement `Config` I have to provide implementations of `getList` and when creating the instance of the concrete implementation I can parameterize it however I need. So maybe `MapConfig` should take a list delimiter, but for things like `CommonsToConfig`, `TypesafeConfig`, or say a `JsonConfig` a list delimiter param is not needed and feels out of place.

howardyuan:
@brharrington Thanks for the heads up on the issue of getList. I believe the issue lies in the way of Archaius1 stores the property. It implicitly split up property into list and stores it as a list. Archaius2 only picks up the head. You can see this by just doing config.getString, we would get "a" instead of "a,b". I think this is a bigger issue than just getList. Whenever the property has listDelimiter in it, it will cause discrepancy between Archaius1 and Archaius2. I'll file an issue and fix it.
Let's talk about API more. As I said before, this is mostly for parity between Archaius1 and Archaius2. To me, if we provide getList in core API, we need to provide a way to set delimiter as well. Also, if we only provide this method in each config impl, I couldn't think of a consistent and convenient way of doing this in CompositeConfig. I thought of providing a static setDelimiter method is AbstractConfig so we don't have to expose the API, but Archaius2 is avoiding static at all cost, so that's a no-go there. My 2c.

brharrington:
> To me, if we provide getList in core API, we need to provide a way to set delimiter as well.

Isn't that assuming that a list is always stored as a string with a simple delimiter? Right now there are two sub-libraries providing bindings for commons configuration and typesafe config (see #266). Both of those already provide constructs for the user to get lists. It seems like archaius2 is going the direction of ignoring that and defining a list is always a string with a simple delimiter. This makes it confusing to use because now if I have a config file I have mapped in I need to ignore the docs for those libraries that tell me how to represent a list and realize that archaius2 will always just get a string and try to parse a list out of that itself.

Further, as a user it isn't a property I could ever safely use unless I can enforce how every property in my system chooses to encode a list as a string. Suppose I have two libraries one assuming commas and one assuming spaces either using annotations as shown below or in property files: 

```java
@Configuration
interface ConfigFromLibraryA {
    @DefaultValue("a,b")
    List<String> foo();
}

@Configuration
interface ConfigFromLibraryB {
    @DefaultValue("a b")
    List<String> foo();
}

// As a side note, it would probably be better to allow the default value to simply be an array if the type
// is a list. In which case this is another example where the setting isn't needed.
@Configuration
interface ConfigFromLibraryC {
    @DefaultValue({"a" , "b"})
    List<String> foo();
}
```

In this case at the very least the parameter should be limited to a specific leaf in the config tree and not assumed to be global for everything in the CompositeConfig or that the setting should be blindly propagated to all sub-configs.

To me the user of `Config` knows there is a property with a value that is a list. That user doesn't necessarily know where the property came from or how it was encoded as a list. So having them set a flag for how to parse a list doesn't make sense.

howardyuan:
@brharrington Thanks again for raising the concerns. I think what we may end up doing is to get rid of the concrete implementations in CompositeConfig and delegate operations like getList etc. to its children in their order. This way, we don't have to expose awkward APIs like setListDelimiter in core config interface.

archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
```diff
@@ -55,4 +59,64 @@ public Object getRawProperty(String key) {
...
+    @Override
+    public <T> List<T> getList(String key, Class<T> type) {
+        List value = config.getList(key);
+        if (value == null) {
+            return notFound(key);
+        }
+;
+        List<T> result = new ArrayList<T>();
+        for (Object part : value) {
+            result.add(getDecoder().decode(type, part.toString()));
```
elandau: No need to call the decoder if part is already of type 'type'

archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
```diff
@@ -55,4 +59,64 @@ public Object getRawProperty(String key) {
...
+    public List getList(String key) {
+        List value = config.getList(key);
+        if (value == null) {
+            return notFound(key);
+        }
+        return value;
+    }
+
+    @Override
+    public String getString(String key, String defaultValue) {
```
elandau: Why the extra code here to handle getString() as a list?  Why not just expect it be a list or convert to a string if the returned object is of type List

archaius2-core/src/main/java/com/netflix/archaius/DefaultDecoder.java
```diff
@@ -42,6 +44,15 @@
             return (T) encoded;
         }
         else if (type.equals(boolean.class) || type.equals(Boolean.class)) {
+            if (type.equals(boolean.class) || type.equals(Boolean.class)) {
```
elandau: Whey the duplicate 'type.equals(boolean.class) || type.equals(Boolean.class'?

archaius2-core/src/main/java/com/netflix/archaius/config/CompositeConfig.java
```diff
@@ -240,6 +241,26 @@ public Object getRawProperty(String key) {
     }
 
     @Override
+    public <T> List<T> getList(String key, Class<T> type) {
+        for (Config child : children) {
+            if (child.containsKey(key)) {
+                return child.getList(key, type);
+            }
+        }
+        return null;
```
elandau: call notFound

archaius2-core/src/main/java/com/netflix/archaius/config/CompositeConfig.java
```diff
@@ -240,6 +241,26 @@ public Object getRawProperty(String key) {
...
+    }
+
+    @Override
+    public List getList(String key) {
+        for (Config child : children) {
+            if (child.containsKey(key)) {
+                return child.getList(key);
+            }
+        }
+        return null;
```
elandau: call notFound instead of returning null

howardyuan:
Addressed your comments. Let me know if you have any questions.

PR #367, opened 2015-11-17
2.x factoring annotations and interfaces into archaius2-api.

qiangdavidliu:
also changing findbugs:annotations to findbugs:jsr305


archaius2-api/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+apply plugin: 'java'
+
+dependencies {
+    compile     'org.slf4j:slf4j-api:1.6.3'
+    compile     'javax.inject:javax.inject:1'
+    compile     'com.google.code.findbugs:jsr305:3.0.1'
+
+    testCompile project(":archaius2-core")
```
elandau: Let's move all the tests into core.  Putting this testCompile dependency here actually causes a circular dependency (archaius2-api -> archaius2-core -> archaius2-api) which confuses eclipse.

qiangdavidliu: that's eclipse's problem:)

archaius2-api/src/main/java/com/netflix/archaius/ConfigLoader.java
```diff
@@ -78,7 +76,7 @@
          * @param resourceName
          * @return CompositeConfig contains a full hierarchy of cascaded files
          */
-        CompositeConfig load(String resourceName) throws ConfigException;
+        Config load(String resourceName) throws ConfigException;
```
elandau: This method returns CompositeConfig because configuration loading is inherently composite due to cascade loading.  Exposing CompositeConfig from the API makes that explicit and also allows for post processing of the response.  With that said, we would need to make CompositeConfig an interface for this to be part of the api package.

PR #369, opened 2015-11-30
Performance optimization for primitive Dynamic***Property classes

kerumai:
This is an optimization to reduce the cpu hotspots I was seeing under load with usage of DynamicBooleanProperty/DynamicIntProperty getValue implementation.
As we use DynamicBooleanProperty's very frequently for feature-flagging in zuul, I was seeing this quite high up when profiling zuul under high load. It seems to be caused by overhead around autoboxing of the value objects to primitives (although I haven't verified this).

I initially added a helper class in zuul for this - https://github.com/Netflix/zuul/blob/2.x/zuul-core/src/main/java/com/netflix/zuul/properties/CachedProperties.java - but would like to apply the same principle in archaius itself, and so have done this PR.

The PR basically changes the Dynamic{Primitive}Property classes to store the cached primitive value, and update it using the callback mechanism. And then change the existing get() method to return this cached value instead of boxing the property value object.

I've included a basic benchmarking class that compares performance of the old and new implementations for boolean. See CachedPropertiesPerfTest.


archaius-core/src/main/java/com/netflix/config/DynamicBooleanProperty.java
```diff
@@ -23,14 +23,41 @@
...
+        });
     }
+
     /**
      * Get the current value from the underlying DynamicProperty
+     *
+     * @return
+     */
+    private boolean chooseValue() {
+        Boolean propValue = this.prop == null ? null : this.prop.getBoolean();
```
pstout: The null check for the prop field is unnecessary.

pstout: This method could be just the previous body of the get method: return prop.getBoolean(defaultValue).booleanValue().  The boxing and unboxing that will occur when the property value changes will be off the hot-path and a minor overhead for clearer code.

archaius-core/src/test/java/com/netflix/config/DynamicPropertyCachedPrimitiveTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        assertEquals(1, prop.get());
+        assertEquals(1, prop.getValue().intValue());
+
+        // Change the value of the property.
+        ConfigurationManager.getConfigInstance().setProperty(propName, "2");
+        assertEquals(2, prop.get());
+        assertEquals(2, prop.getValue().intValue());
+
+        // Remove the prop value, should change to default of false.
```
pstout: nit: false --> zero

archaius-core/src/test/java/com/netflix/config/DynamicPropertyCachedPrimitiveTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        assertEquals(1, prop.get());
+        assertEquals(1, prop.getValue().longValue());
+
+        // Change the value of the property.
+        ConfigurationManager.getConfigInstance().setProperty(propName, "2");
+        assertEquals(2, prop.get());
+        assertEquals(2, prop.getValue().longValue());
+
+        // Remove the prop value, should change to default of false.
```
pstout: nit: false --> zero

archaius-core/src/test/java/com/netflix/config/DynamicPropertyCachedPrimitiveTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        assertEquals(1.1d, prop.get());
+        assertEquals(1.1d, prop.getValue().doubleValue());
+
+        // Change the value of the property.
+        ConfigurationManager.getConfigInstance().setProperty(propName, "2.2");
+        assertEquals(2.2d, prop.get());
+        assertEquals(2.2d, prop.getValue().doubleValue());
+
+        // Remove the prop value, should change to default of false.
```
pstout: nit: false --> 0.5

archaius-core/src/test/java/com/netflix/config/DynamicPropertyCachedPrimitiveTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        assertEquals(1.1f, prop.get());
+        assertEquals(1.1f, prop.getValue().floatValue());
+
+        // Change the value of the property.
+        ConfigurationManager.getConfigInstance().setProperty(propName, "2.2");
+        assertEquals(2.2f, prop.get());
+        assertEquals(2.2f, prop.getValue().floatValue());
+
+        // Remove the prop value, should change to default of false.
```
pstout: nit: false --> 0.5

archaius-core/src/main/java/com/netflix/config/DynamicBooleanProperty.java
```diff
@@ -23,14 +23,41 @@
...
+    protected volatile boolean primitiveValue;
+
     public DynamicBooleanProperty(String propName, boolean defaultValue) {
         super(propName, Boolean.valueOf(defaultValue));
+
+        // Set the initial value of the cached primitive value.
+        this.primitiveValue = chooseValue();
+
+        // Add a callback to update the cached primitive value when the property is changed.
+        this.prop.addCallback(new Runnable() {
```
pstout: Rather than adding this callback here, you could just override either of the propertyChanged methods and then update the PropertyWrapper class to not add this class to the SUBCLASSES_WITH_NO_CALLBACK map.  This would be cleaner and probably result in less overall code.

archaius-core/src/main/java/com/netflix/config/DynamicBooleanProperty.java
kerumai: Actually, all of the overhead does seem to be caused by this. I've added another benchmark comparison to the CachedPropertiesPerfTest to illustrate. eg:

    Original DynamiceBooleanProperty totalled 7087 ms.
    New DynamicBooleanProperty totalled 1252 ms.
    New DynamicBooleanProperty with old get() totalled 7129 ms.

archaius-core/src/main/java/com/netflix/config/DynamicBooleanProperty.java
kerumai: Yes that is a much better way of doing it. 
@howardyuan - i've made this change to the PR. Does this satisfy your concerns around my previous usage of the callbacks?

archaius-core/src/main/java/com/netflix/config/DynamicBooleanProperty.java
pstout: I wasn't suggesting reverting the change to the get method, which is what your new test is effectively doing, but rather to put the old code from the get method into the new chooseValue method.  There it would be execute just once per property value change.

kerumai: Ah I see what you mean now. Sorry for the confusion. I'll do that now, thanks.

pstout:
LGTM

archaius-core/src/main/java/com/netflix/config/DynamicBooleanProperty.java
```diff
@@ -23,15 +23,42 @@
...
     public DynamicBooleanProperty(String propName, boolean defaultValue) {
         super(propName, Boolean.valueOf(defaultValue));
+
+        // Set the initial value of the cached primitive value.
+        this.primitiveValue = chooseValue();
+    }
+
+    @Override
+    protected void propertyChanged() {
+        super.propertyChanged();
```
howardyuan: I'm a little wary of calling super here. It's a no-op there at super, so overhead is less of concern. But it maybe confusing for people down the road who might expect if they make the change in super, it would take effect in subclasses as well.

archaius-core/src/main/java/com/netflix/config/PropertyWrapper.java
```diff
@@ -41,12 +41,7 @@
     private final List<Runnable> callbackList = Lists.newArrayList();
 
     static {
-        PropertyWrapper.registerSubClassWithNoCallback(DynamicIntProperty.class);
```
howardyuan: So this still incurs the overhead of callbacks both in terms of memory and invocation. Any reason that we don't want to introduce cached properties so people are aware of the cost and can make the choice? I'm concerned people will be surprised by the behavior changes.

kerumai: Although creating alternate cached properties would work, I'm concerned that few teams will make the change to use them. And for large systems with many libraries using the existing properties, it ideally would be the choice of the system owners to change this globally for their systems, rather than have to get each library owner to change it in turn.

Taking a look at a large system in prod, jmap shows me ~8000 instances of Dynamic(Boolean|Long|Int|Float|Double)PropertyS, which if 1 instance of a callback conservatively adds 32 bytes, would imply a memory increase of only 256 kB. This seems a reasonable overhead increase to me.

And the overhead of invoking the callback only happens when a property value is changed, which does not happen frequently, and does not happen on the critical path.

So it's ultimately you call, but I think pro's outweigh the cons for making this change to the existing impls.

archaius-core/src/main/java/com/netflix/config/DynamicBooleanProperty.java
kerumai: Yeah thats true. Will change that now.

archaius-core/src/main/java/com/netflix/config/DynamicDoubleProperty.java
```diff
@@ -39,4 +39,4 @@ public double get() {
     public Double getValue() {
         return get();
     }
-}
+}
```
howardyuan: what're all these format changes? can we get rid of those format changes only files?

kerumai: Absolutely.

PR #372, opened 2015-12-16
Backport from Karyon

elandau:
* Simplify ArchaiusModule to make it less dependent on being customized using DI
* Remove the concept of library overrides provided as guice bindings (it's overly complex and not necessary)
* Get rid of the ArchaiusConfiguration class
* Get rid of the ConfigSeeder concept as it isn't necessary
* Fix ConfigProxyFactory's use of the deprecated ProxyFactory


archaius2-core/src/main/java/com/netflix/archaius/ProxyFactory.java
```diff
@@ -30,267 +16,10 @@
...
 @Singleton
 @Deprecated
-public class ProxyFactory {
-    
-    /**
-     * The decoder is used for the purpose of decoding any @DefaultValue annotation
-     */
-    private final Decoder decoder;
-    private final PropertyFactory propertyFactory;
+public class ProxyFactory extends ConfigProxyFactory {
```
brharrington: What is the rationale for keeping this class around?

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,261 @@
...
+import com.netflix.archaius.config.EnvironmentConfig;
+import com.netflix.archaius.config.MapConfig;
+import com.netflix.archaius.config.SystemConfig;
 import com.netflix.archaius.interpolate.ConfigStrLookup;
 import com.netflix.archaius.readers.PropertiesConfigReader;
 
 /**
- * Guice module with default bindings to enable Config injection and 
- * configuration mapping/binding in a Guice based application.
+ * Guice Module for setting up archaius and making it's components injectable and provides the following
```
brharrington: s/it's/its/

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,261 @@
...
+ * functionality
+ * 
+ * <ul>
+ * <li>Configuration Proxy</li>
+ * <li>Injectable Config</li>
+ * <li>Configuration mapping</li>
+ * </uL>
+ * 
+ * Note that this module has state and should therefore only be installed once.  It should only
+ * be installed by applications and not by libraries depending on archaius that.  Such libraries
```
brharrington: s/ that//

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,261 @@
...
+ * 
+ * <ul>
+ * <li>Configuration Proxy</li>
+ * <li>Injectable Config</li>
+ * <li>Configuration mapping</li>
+ * </uL>
+ * 
+ * Note that this module has state and should therefore only be installed once.  It should only
+ * be installed by applications and not by libraries depending on archaius that.  Such libraries 
+ * should only specify a bindger.requriesBinding(Config.class).
```
brharrington: s/bindger.requriesBinding/binder.requiresBinding/

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,261 @@
...
+    private static final String REMOTE_LAYER_NAME       = "REMOTE";
+    private static final String SYSTEM_LAYER_NAME       = "SYSTEM";
+    private static final String ENVIRONMENT_LAYER_NAME  = "ENVIRONMENT";
+    private static final String APPLICATION_LAYER_NAME  = "APPLICATION";
+    private static final String LIBRARIES_LAYER_NAME    = "LIBRARIES";
     
-    private static final AtomicInteger idCounter = new AtomicInteger();
+    static {
+        System.setProperty("archaius.default.configuration.class",      "com.netflix.archaius.bridge.StaticAbstractConfiguration");
+        System.setProperty("archaius.default.deploymentContext.class",  "com.netflix.archaius.bridge.StaticDeploymentContext");
```
brharrington: Shouldn't this be done in the bridge?

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,261 @@
...
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
 
-    @Override
-    public int hashCode() {
-        return ArchaiusModule.class.hashCode();
+        throw new ProvisionException("Only one ArchaiusModule may be installed");
```
brharrington: Having equals throw to enforce a singleton module policy seems unintuitive. Does this really add much over the normal duplicate binding for Config error?

archaius2-core/src/main/java/com/netflix/archaius/ProxyFactory.java
elandau: I suppose it's been deprecated long enough.  I'll delete it.

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
elandau: Yes, but there were some static initialization ordering issues with code that was previously invoked as part of archaius2 bootstrapping which triggered the wrong classes being created in archaius1.

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
elandau: I felt that this exception may give more details about the root cause than a duplicate binding error.  However, I also agree that it doesn't make sense to throw an exception from equals().  I'll fix this.

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,251 @@
...
-    @DefaultsLayer
-    SettableConfig getDefaultsConfig() {
-        return new DefaultSettableConfig();
+        return runtimeLayer;
     }
     
     @Provides
     @Singleton
-    @ApplicationLayer 
-    CompositeConfig getApplicationLayer() {
```
howardyuan: So there's no getter for Application layer and Remote layer? Any reason for that?

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,251 @@
...
-            @DefaultsLayer            SettableConfig    defaultsLayer) throws ConfigException {
-        Builder builder = DefaultCompositeConfig.builder()
-                .withConfig(RUNTIME_LAYER_NAME,              settableLayer)
-                .withConfig(REMOTE_LAYER_NAME,               overrideLayer)
-                .withConfig(SYSTEM_LAYER_NAME,               SystemConfig.INSTANCE)
-                .withConfig(ENVIRONMENT_LAYER_NAME,          EnvironmentConfig.INSTANCE);
-        
-        if (archaiusConfiguration.getApplicationOverride() != null) {
-            builder.withConfig(APPLICATION_OVERRIDE_LAYER_NAME, archaiusConfiguration.getApplicationOverride());
+    Config getConfig(ConfigLoader loader, Injector injector) throws Exception {
```
howardyuan: Does this mean that you always need a bootstrap injector for Archaius module to work right?

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,251 @@
...
+        if (applicationOverrides != null) {
```
howardyuan: I'm a bit puzzled by the approach here. So the only way to apply overrides is through this singleton config? So people can't control the overrides through the code?

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,251 @@
...
         
-        // First load the single application configuration 
-        CompositeConfig compositeConfig = loader
-                .newLoader()
-                .withCascadeStrategy(NoCascadeStrategy.INSTANCE)
-                .load(archaiusConfiguration.getConfigName());
-        if (compositeConfig != null) {
-            applicationLayer.replaceConfig(APPLICATION_LAYER_NAME, compositeConfig);
-        }
+        this.applicationLayer.addConfig("loaded",  loader
```
howardyuan: This "loaded" layer is added after overrides? Is the sequence reversed?

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -15,299 +15,251 @@
...
-            runtimeLayer.setProperties(provider.get(config));
-        }
- 
-        // Finally, load any cascaded configuration files for the application
-        compositeConfig = loader.newLoader().withCascadeStrategy(archaiusConfiguration.getCascadeStrategy()).load(archaiusConfiguration.getConfigName());
-        if (compositeConfig != null) {
-            applicationLayer.replaceConfig(APPLICATION_LAYER_NAME, compositeConfig);
+        Binding<Config> binding = injector.getExistingBinding(Key.get(Config.class, RemoteLayer.class));
+        if (binding != null) {
+            // TODO: Ideally this should replace the remoteLayer in config but there is a bug in archaius
```
howardyuan: Hmm, what's is this bug? I'm not aware of it?

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
elandau: I'm trying to reduce the API so it only exposes things are should be consumed.  I don't think the @ApplicationLayer is something that should be consumed.

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
elandau: No, it's the current injector being injected.  I'll clean this up a bit so we don't inject the injector.

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
elandau: You can set overrides in code by injecting 

@RuntimeLayer
SettableConfig config;

Issue #373, opened 2015-12-18
EtcdConfigurationSource doesn't support "update" action


dborovikov:
If I add callback to existing in etcd property and change value using PUT command callback is not called. Callback is called only if I create new key ("set" command). To support property updates "update" command should be also implemented.

rgallardo-netflix:
V1 is deprecated and won't be maintained anymore.

PR #374, opened 2015-12-18
Fix for https://github.com/Netflix/archaius/issues/373

dborovikov:


archaius-etcd/src/main/java/com/netflix/config/source/EtcdConfigurationSource.java
```diff
@@ -129,6 +129,11 @@ public void handle(Response updateResponse) {
                         updateConfiguration(createIncremental(null, ImmutableMap.<String, Object>of(sourceKey, value), null));
                         break;
 
+                    case "update":
```
howardyuan: Do you want to check first if sourceKey exists when doing "Update"? If not, you can either throws some exception or change to Set? I'm not familiar with etcd module. I'll leave the PR open for now and hopefully @spoon16 can take a look. Also, I probably won't be able to release a new Archaius until mid Jan.

rgallardo-netflix:

PR #376, opened 2016-01-05
Cleanup PropertyFactory API

elandau:
Clean up the PropertyFactory and Property APIs to decouple them completely from the Config API so they can be backed by other configuration mechanisms.


howardyuan:
LGTM

PR #377, opened 2016-01-06
README updates, tests and code clean up on 2.x

drtechniko:
Cleaned up some unused code and added unit tests after running local code coverage. Please check if the code I deleted is indeed not used. Other than doing a search within the project and in Stash across projects I didn't have a good way of checking.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/DefaultConfigLoader.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/DefaultConfigLoader.java
@@ -18,10 +18,7 @@ package com.netflix.archaius;
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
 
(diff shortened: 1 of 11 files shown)


drtechniko:
@elandau and @howardyuan Tests look good. Please have a look and let me know.

archaius2-core/src/main/java/com/netflix/archaius/DefaultConfigLoader.java
```diff
@@ -70,7 +70,13 @@ public Builder withConfigReader(ConfigReader loader) {
             this.loaders.add(loader);
             return this;
         }
-        
+
+        public Builder withConfigReaders(Set<ConfigReader> loaders) {
```
brharrington: Why does this take a `Set`? Since `this.loaders` is a List multiple calls can still result in duplicates.

elandau: this.loaders should be a Set

drtechniko: Fixed with the new commit.
If all else is cool I will merge.

PR #380, opened 2016-01-14
Re-introduce DefaultLayer that can be populated via bindings

elandau:


archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -155,8 +162,7 @@ public ArchaiusModule withApplicationOverrides(Properties prop) {
     
     public ArchaiusModule withApplicationOverrides(Config config) {
         try {
-            uniqueNameCounter++;
-            applicationLayer.addConfig("override-" + uniqueNameCounter, config);
+            applicationLayer.addConfig(getUniqueName("override-"), config);
```
qiangdavidliu: delete the "-" if there is no longer a trailing counter?

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -164,17 +170,15 @@ public ArchaiusModule withApplicationOverrides(Config config) {
         return this;
     }
     
-    public ArchaiusModule withCascadeStrategy(Class<? extends CascadeStrategy> cascadeStrategy) {
```
qiangdavidliu: Public api removal. Is is safe?

PR #382, opened 2016-01-22
Final fix for ArchaiusModule Guice best practices

elandau:


archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -150,169 +112,143 @@ public ArchaiusModule withCascadeStrategy(Class<? extends CascadeStrategy> casca
         return this;
     }
     
-    private String getUniqueName(String prefix) {
-        uniqueNameCounter++;
-        return prefix +"-" + uniqueNameCounter;
+    protected void configureArchaius() {
```
brharrington: I thought there was a recommendation to discourage sub-classes and if possible make the module final. In particular, how does the deduping still work if a module is registered along with a subclass of that module?

elandau: This follows the same approach as ServletModule where sub-classing the module is only done for customization purposes.  Note that InternalArchaiusModule is final and has the proper deduping equals/hashCode.

archaius2-guice/src/main/java/com/netflix/archaius/guice/ArchaiusModule.java
```diff
@@ -121,21 +77,27 @@
         System.setProperty("archaius.default.deploymentContext.class",  "com.netflix.archaius.bridge.StaticDeploymentContext");
     }
     
-    private String configName = DEFAULT_CONFIG_NAME;
-    
+    @Deprecated
```
qiangdavidliu: This is a private field, so we can just delete it instead of deprecating.

PR #384, opened 2016-01-25
Fix equals for InternalArchaiusModule

qiangdavidliu:


elandau:
ArchaiusModule should not be dedupped as it is only used for customizing archaius.  You should be able to install multiple of these, each with different customizations.  This follows the approach used by ServletModule

PR #387, opened 2016-02-02
Move to the current version of type safe config 

gorzell:
and make the versioning more flexible for minor releases.


gorzell:
Looks like they build with java 8.  Is there an option to move that direction before release?

PR #388, opened 2016-02-03
'Fix' default value in interpolation

elandau:
Modify syntax for interpolation with default value to have just the standard ':' separator instead of ':-'


brharrington:
The `InterpolationTest` added doesn't seem to cover the actual change. If I understand correctly, it is changing `${foo:-DEFAULT}` to be `${foo:DEFAULT}`.

What is meant by the standard `:`? Is that syntax used somewhere else? The `:-` pattern is more familiar to me as it is used by bash for default values as part of parameter expansion:

http://wiki.bash-hackers.org/syntax/pe#use_a_default_value

spencergibb:
It's the standard for Spring, if that matters.

elandau:
I'll see if I can add support for both but the plain ':' is a much more familiar syntax for specifying defaults, especially to anyone who has used Spring.  

Regarding the unit tests, they were somehow removed when I was doing some cleanup.  I'll re-add them.

brharrington:
Thanks for the explanation. I'm not opposed to the change, just curious about the motivation. There is no reason to try and support multiple ways of specifying defaults on my account. I think it is better overall to have a single way as it reduces the amount of stuff a user would need to learn.

elandau:
@brharrington Thanks for the input.  After take a very short look at supporting both I realized that it really makes no sense to attempt that as it would cause ambiguity when dealing with negative numbers.

Issue #390, opened 2016-02-08
Shorthand for adding properties file in ArchaiusModule


twicksell:
It would be nice to be able to easily specify additional properties when creating an Injector. Something along the lines of:

```
InjectorBuilder.fromModules(new SomeOtherModule(), new ArchaiusModule() {
      @Override
       protected void configureArchaius() {
                 addConfigFromPropertiesFile("foo.properties");
       }
});
```

howardyuan:
As of current Archaius2, the override through any resource can be achieved as following,
            @Override
            protected void configureArchaius() {
                try {
                    bindApplicationConfigurationOverride().toInstance(
                            DefaultConfigLoader.builder().build().newLoader().load("application-override"));
                } catch (ConfigException e) {
                    throw new RuntimeException(e);
                }
            }
        });

elandau:
Solved with #411

PR #391, opened 2016-02-09
Implementation for #390

elandau:
Implementation for #390.

Add PropertiesConfig to load from properties file.   ArchaiusModule shorthand for adding overrides from a file

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/config/PropertiesConfig.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/config/PropertiesConfig.java
@@ -1,0 +1,23 @@
+package com.netflix.archaius.config;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+public class PropertiesConfig extends MapConfig {
+
+    public static PropertiesConfig fromFile(String filename) throws IOException {
+        Properties prop = new Properties();
+        try (InputStream in = Thread.currentThread().getContextClassLoader().getResourceAsStream(filename)) {
+            if (in == null) 
+                throw new IOException("Resource " + filename + " not found");
+            prop.load(in);
+            return new PropertiesConfig(prop);
+        }
+    }
+    
+    public PropertiesConfig(Properties props) {
+        super(props);
+    }
+
+}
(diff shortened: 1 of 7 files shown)


archaius2-core/src/main/java/com/netflix/archaius/config/PropertiesConfig.java
```diff
@@ -0,0 +1,23 @@
+package com.netflix.archaius.config;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+public class PropertiesConfig extends MapConfig {
+
+    public static PropertiesConfig fromFile(String filename) throws IOException {
```
brharrington: Based on the name, I would expect this to load from the file system rather than a class loader. Perhaps change to `fromResource` instead?

It would also be useful to have a comment indicating the behavior if multiple resources with the provided name are in the classpath. For example, in archaius v1 the helpers would use `getResources` and load all of the matching property files. In this case it only gets the first one that is found.

elandau: @brharrington Thanks for the input.  I could modify the implementation to consider classpath as well as file system.  Would that work?  

What, in your opinion, should be the expected behavior if there are multiple matching files?  Should it fail with a multiple files exception, should it load all, should it use the first?

brharrington: > I could modify the implementation to consider classpath as well as file system.

In most of the APIs I have seen, files vs resources are separate and the user explicitly chooses. So if both are needed I would expect two methods, something like:

* `fromFile`
* `fromResources`

Examples from other libraries, guava has both Resources and Files helpers:

http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/io/Resources.html
http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/io/Files.html

Typesafe config has separate methods on the factory:

http://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigFactory.html

> What, in your opinion, should be the expected behavior if there are multiple matching files?

The others I have looked at archaiusv1 and [typesafe config](http://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigFactory.html#parseResources-java.lang.Class-java.lang.String-com.typesafe.config.ConfigParseOptions-) both merge. I'm not sure of the override behavior for archaiusv1, but the typesafe config the first encountered has priority. However, there is no guarantee of the order they will be found in scanning the classpath.

I would probably choose to fail or load all and merge. The main argument for merge is that it seems to be more consistent with other libraries. It is just a helper so the user can always implement their own method to load the properties object if they need custom behavior. I think the most important thing is that the behavior is clearly indicated to the user.

twicksell:
Resolved in https://github.com/Netflix/archaius/pull/411

PR #392, opened 2016-02-25
Preserve order or elements in DynamicSetProperty and DynamicMapProperty

elandau:


archaius-core/src/test/java/com/netflix/config/DynamicMapPropertyTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+        @Override
+        protected Integer from(String value) {
+            return Integer.parseInt(value);
+        }
+    }
+    
+
+    @Test
+    public void testCallbacksAddUnsubscribe() {
```
howardyuan: For test cases, can you change the initial set value to an unordered list so we can verify the new map, set property actually preserves the "random" order?

elandau: I'll mess with the order a bit

PR #393, opened 2016-03-09
add junit rule Archaius2TestConfig to ease property overrides in testing

twicksell:


archaius2-test/src/test/java/com/netflix/archaius/test/Archaius2RuleTest.java
```diff
@@ -0,0 +1,47 @@
+package com.netflix.archaius.test;
+
+import static org.junit.Assert.*;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+@TestPropertyOverride({"typeLevelProperty=present", "cascading=type"})
```
howardyuan: typo? topLevel?

archaius2-test/src/test/java/com/netflix/archaius/test/Archaius2RuleTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Rule
+    public Archaius2TestConfig config = new Archaius2TestConfig();
+    
+    @Test 
+    @TestPropertyOverride({"testLevelProperty=present"})
+    public void testBasicPropertyResolution() {
+        assertNotNull(config);
+        assertEquals("present", config.getString("parentClassLevelProperty"));
+        assertEquals("present", config.getString("classLevelProperty"));
+        assertEquals("present", config.getString("testLevelProperty"));
```
pstout: Consider adding an assert to verify that the value of cascading is "type".  You never actually verify that multiple property values are parsed correctly. :-/

archaius2-test/src/test/resources/archaiusRuleTest.properties
```diff
@@ -0,0 +1 @@
+fileProperty=present
```
pstout: Are you intentionally testing that the last property loads when there is no trailing new line?  If so, I would add a comment explaining that fact, otherwise consider adding a newline to get rid of the angry red icons.

archaius2-test/src/main/java/com/netflix/archaius/test/TestPropertyOverride.java
```diff
@@ -0,0 +1,45 @@
...
+     * ex. {@literal @}TestPropertyOverride({"propName=propValue", "propName2=propVal2", ... }) 
+     * 
+     * These properties will precedance over those created from the
+     * propertyFiles attribute in the event of conflicts. 
+     */
+    String[] value() default {};
+    
+    /***
+     * Use a file location to create properties.
+     * ex. {@literal @}TestPropertyOverride(locations={"unittest.properties"})
```
pstout: Looks like the Javadoc has drifted from the actual name of the attribute.

archaius2-test/src/main/java/com/netflix/archaius/test/Archaius2TestConfig.java
```diff
@@ -0,0 +1,289 @@
...
+ * See {@link TestPropertyOverride} for additional usage information.
+ */
+public class Archaius2TestConfig implements TestRule, Config {
+    private final ThreadLocal<Config> configHolder = new ThreadLocal<>();
+
+    public Statement apply(Statement base, Description description) {
+        return statement(base, description);
+    }
+
+    private Statement statement(final Statement base, final Description description) {
```
pstout: I don't understand the value of extracting this method from the public apply method.  Why just mark the parameters of that method final and remove lines 42-45?

twicksell: The apply method is coming from the TestRule interface, where unfortunately the parameters are not final.

pstout: The finality of arguments is specific to the method implementation.  It does not matter how the interface declares them.

PR #395, opened 2016-03-24
Fix property caching

elandau:
Fix DefaultPropertyContainer so that it actually caches the new value and version stamp

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/property/DefaultPropertyContainer.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/property/DefaultPropertyContainer.java
@@ -174,35 +174,6 @@ public class DefaultPropertyContainer implements PropertyContainer {
         }
         
         /**
-         * Update to the latest value and return either the new value or previous value if not updated
-         * or failed to resolve.
-         * 
-         * @param latestVersion
-         * @return
-         */
-        T updateAndGet(int latestVersion) {
-            int[] cacheVersion = new int[1];
-            T existing = cache.get(cacheVersion);
-            try {
-                T value = resolveCurrent();
-                if (value != existing && (value == null || existing == null || !existing.equals(value))) {
-                    if (cache.compareAndSet(existing, value, cacheVersion[0], latestVersion)) {
-                        // Slight race condition here but not important enough to warrent locking
-                        lastUpdateTimeInMillis = System.currentTimeMillis();
-                        return value;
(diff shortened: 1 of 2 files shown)


archaius2-core/src/test/java/com/netflix/archaius/DefaultPropertyContainerTest.java
```diff
@@ -0,0 +1,22 @@
...
+    @Test
+    public void basicTest() {
+        SettableConfig config = new DefaultSettableConfig();
+        DefaultPropertyFactory factory = DefaultPropertyFactory.from(config);
+        Property<String> prop = factory.getProperty("foo").asString("default");
+
+        Assert.assertEquals("default", prop.get());
+        config.setProperty("foo", "value1");
+        Assert.assertEquals("value1", prop.get());
+        Assert.assertEquals("value1", prop.get());
```
howardyuan: How do we know the second one uses cache and the first one not?

archaius2-core/src/main/java/com/netflix/archaius/property/DefaultPropertyContainer.java
```diff
@@ -174,35 +174,6 @@ public void removeListener(PropertyListener<T> listener) {
...
-         * 
-         * @param latestVersion
-         * @return
-         */
-        T updateAndGet(int latestVersion) {
-            int[] cacheVersion = new int[1];
-            T existing = cache.get(cacheVersion);
-            try {
-                T value = resolveCurrent();
-                if (value != existing && (value == null || existing == null || !existing.equals(value))) {
```
howardyuan: so the bug here is that if values are the same before current and cached, the version will not be updated?

archaius2-core/src/test/java/com/netflix/archaius/DefaultPropertyContainerTest.java
elandau: We currently don't.  The API would need to be re-worked to allow visibility into that.  For now I just stepped through it :(

archaius2-core/src/main/java/com/netflix/archaius/property/DefaultPropertyContainer.java
elandau: There were a few other use cases but essentially yes.  If the value was anything but new the version timestamp never got updated.

PR #396, opened 2016-03-25
Fix interpolation of string in @DefaultValue

elandau:


archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -256,7 +259,13 @@ public T get() {
     }
 
     private <T> MethodInvoker<T> createDynamicProperty(final Class<T> type, final String propName, final String defaultValue) {
-        final Property<T> prop = propertyFactory.getProperty(propName).asType(type, defaultValue != null ? decoder.decode(type, defaultValue) : null);
+        final Property<T> prop = propertyFactory
+                .getProperty(propName)
+                .asType(type, defaultValue != null 
+                    // This is a hack to force interpolation of the defaultValue assuming
+                    // that ther is never a property '*'
+                    ? decoder.decode(type, config.getString("*", defaultValue))
```
howardyuan: I'll dig into this later. But this change only covers ConfigProxyFactory case, what about other code paths that needs to interpolate the default value? That's I think it may makes sense to do it in a lower level (inside Config themselves, e.g.?)

PR #398, opened 2016-03-29
Fix non-string dynamic property bridge

elandau:
Force getString() to do a .toString() since all the dynamic configuration classes depend on properties being stored as strings


howardyuan:
LGTM

jhspaybar:
:+1:

PR #399, opened 2016-03-30
Fix broken interpolation in the archaius1-archaius2 bridge

elandau:


archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -128,8 +128,7 @@ public String getString(String key, String defaultValue) {
         if (value instanceof String) {
             return interpolator.create(getLookup()).resolve(value.toString());
         } else {
-            throw new UnsupportedOperationException(
-                    "Property values other than String not supported");
+            return value.toString();
```
qiangdavidliu: Is this the right contract here? If the type is not a string, should the return be value.toString(), or just defaultValue?

elandau: Yes, it is.

archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/StaticAbstractConfiguration.java
```diff
@@ -80,8 +80,11 @@ public boolean containsKey(String key) {
 
     @Override
     public String getString(String key, String defaultValue) {
-        Object value = getProperty(key);
-        return value != null ? value.toString() : defaultValue;
+        if (delegate == null) {
```
howardyuan: The new change is essentially the same as the previous logic?

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -143,8 +142,7 @@ public String getString(String key) {
...
```
howardyuan: I'm a little wary of this. The config has explicit types getters. If we break the type check here, we lose the benefits of property types?

archaius2-archaius1-bridge/src/main/java/com/netflix/archaius/bridge/StaticAbstractConfiguration.java
elandau: The fix for getting non-string types as string has been pushed down to the archaius Config so it can do the interpolation

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
elandau: Getters for specific types continue to work as before.  This just ensures that you can get a string representation of a property that is not stored as a string.

PR #400, opened 2016-03-30
Enforce max number of retries when doing the initial dynamic configur…

howardyuan:
…ation loading.

This bug manifests itself when bootstrap for the first time. No backend dynamic properties server is available and the archaius inside the server tries to connect to server and always get an exception.


archaius2-persisted2/src/main/java/com/netflix/archaius/persisted2/Persisted2ClientConfig.java
```diff
@@ -46,4 +46,12 @@
...
      */
     String getServiceUrl();
+
+    /**
+     * When set to true, the initial loading of persistent properties has to be successful. Otherwise,
+     * the client will keep re-trying and Archaius initialization will not finish. In turn, the application's
+     * initialization will not finish and application will not start up. This is going to be the default
+     * behavior of persistent client.
+     */
+    boolean getSyncInit();
```
elandau: Now that we're exposing this via configuration let's think of a better name.  How about something 'blockUntilInitialized'?

archaius2-core/src/main/java/com/netflix/archaius/config/polling/FixedPollingStrategy.java
```diff
@@ -44,17 +45,23 @@ public FixedPollingStrategy(long interval, TimeUnit units, boolean syncInit) {
...
         this.units    = units;
         this.syncInit = syncInit;
+
     }
     
     @Override
     public Future<?> execute(final Runnable callback) {
-        while (syncInit) {
+        int numRetries = MAX_NUM_RETRIES;
+        while (true) {
```
elandau: This look looks a bit awkward in how the numRetries update and condition check are disconnected.  This might read better as
```java
while (syncInit) {
   try { 
      callback.run();
      break;
   }
   catch (Exception e) {
      if (--numRetries == 0) {
          return Futures.immediateFailure(e); 
      }
      try {
          units.sleep(interval);
      } catch (InterruptedException e1) {
          Thread.currentThread().interrupt();
          return Futures.immediateFailure(e);
      }
   }
}
```

archaius2-core/src/main/java/com/netflix/archaius/config/polling/FixedPollingStrategy.java
```diff
@@ -29,32 +29,28 @@
...
     private final TimeUnit units;
-    private final boolean syncInit;
-    
+
+
     public FixedPollingStrategy(long interval, TimeUnit units) {
-        this(interval, units, true);
-    }
-    
-    public FixedPollingStrategy(long interval, TimeUnit units, boolean syncInit) {
```
elandau: Since this is a breaking change let's deprecate this constructor instead and log a warning if syntInit is false.

PR #401, opened 2016-04-02
Back port module to enable ConfigProxyFactory for legacy archaius

elandau:
This change makes it possible to use `ConfigProxyFactory` with any application that still uses Archaius1 for its configuration.  Libraries can now develop to the Archaius2 style of configuration using interfaces and `ConfigProxyFactory` to bind those interfaces to configuration.  

To enable just install `Archaius2BackportModule` on the injector and add configuration interface bindings as you would with Archaius2.  Note that Archacius2Backport is meant to be installed instead of `ArchaiusModule` and `StaticArchaiusBridgeModule`.

Note that there are a few minor restrictions here.  This change does not add support for `@ConfigurationSource` annotations in applications still using Archaius1.  For internal Netflix applications configuration classes would need to specify both `@ConfigurationSource` (for Archaius2) and `@NFProperties` (for Archaius1).  @Configuration is also not supported so library modules should specify the property prefix when creating a proxy instance.  

For example,

``` java
// This is the library's configuration interface
@ConfigurationSource("myconfig")
@NFProperties("myconfig")
public interface MyConfig {
    int getSomeProperty();
}
```

```
// This is the library module that an application will need to install
public class MyConfigModule {
    @Override
    public void configure() {}

    @Provides
    @Singleton
    MyConfig getMyConfig(ConfigProxyFactory factory) {
        return factory.newProxy(MyConfig.class, "prefix");
    }
}
```

The above setup will load properties from 'myconfig.properties' into archaius and bind the properties

```
prefix.someProperty=123
```

### DIFF
--- a/archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
+++ b/archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
@@ -16,14 +16,15 @@
 package com.netflix.archaius.commons;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 
 import org.apache.commons.configuration.AbstractConfiguration;
+import org.apache.commons.configuration.event.ConfigurationEvent;
+import org.apache.commons.configuration.event.ConfigurationListener;
+import org.apache.commons.lang3.StringUtils;
 
 import com.netflix.archaius.config.AbstractConfig;
-import org.apache.commons.lang.StringUtils;
 
 /**
  * Adaptor to allow an Apache Commons Configuration AbstractConfig to be used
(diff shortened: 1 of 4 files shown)


archaius2-commons-configuration/src/main/java/com/netflix/archaius/commons/CommonsToConfig.java
```diff
@@ -38,6 +39,12 @@
     
     public CommonsToConfig(AbstractConfiguration config) {
         this.config = config;
+        config.addConfigurationListener(new ConfigurationListener() {
```
howardyuan: I'm really having hard time wrapping my head around all these re-directions. So we're adding a listener here to make sure whoever is depending on CommonsToConfig will have a chance to get notified when underneath config (Archaius1 AbstractConfiguration) changes? Only use case I can think of is one Archaius2 based library being based on another Archaius1 bridged Archaius2 library instead of just directly on the base Archaius1 config? Let me know if I'm wrong, but otherwise, I think we're overloading this bridge. The bridge should just bridge one Archaius1 AbstractConfiguration to another Archaius2 AbstractConfig.

rgallardo-netflix:
We are no longer adding any kind of support for legacy usage.

Issue #403, opened 2016-04-06
Support for typesafe java.time.Duration


costimuraru:
Sample of `reference.conf` file:
`foo = 300ms`

We should be able to map this to a `Duration` attribute.

Exception thrown:
```
WARN com.netflix.governator.configuration.ArchaiusConfigurationProvider - Could not deserialize configuration with key foo to type class java.time.Duration
com.fasterxml.jackson.core.JsonParseException: Unexpected character ('m' (code 109)): Expected space separating root-level values
 at [Source: 300ms; line: 1, column: 5]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1419)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:508)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportUnexpectedChar(ParserMinimalBase.java:437)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportMissingRootWS(ParserMinimalBase.java:453)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._verifyRootSpace(ReaderBasedJsonParser.java:1236)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._parsePosNumber(ReaderBasedJsonParser.java:886)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextToken(ReaderBasedJsonParser.java:680)
	at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:3090)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:3036)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2146)
	at com.netflix.governator.configuration.AbstractObjectConfigurationProvider$1.get(AbstractObjectConfigurationProvider.java:45)
	at com.netflix.governator.configuration.Property$3.get(Property.java:28)
	at com.netflix.governator.lifecycle.ConfigurationProcessor.assignConfiguration(ConfigurationProcessor.java:115)
	at com.netflix.governator.lifecycle.DefaultConfigurationMapper.mapConfiguration(DefaultConfigurationMapper.java:45)
	at com.netflix.governator.lifecycle.LifecycleManager.startInstance(LifecycleManager.java:333)
[...]

PR #405, opened 2016-04-13
Support dynamic configs of different kinds, not just String

costimuraru:
This is very useful for the remote TypesafeConfigLoader, which I’m going to add in a different pull request. It makes it possible to enhance the config with objects of type `Duration`, `Date` and others.

Relates to https://github.com/Netflix/archaius/issues/403

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/readers/PropertiesConfigReader.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/readers/PropertiesConfigReader.java
@@ -64,13 +64,13 @@ public class PropertiesConfigReader implements ConfigReader {
             try {
                 // Load properties into the single Properties object overriding any property
                 // that may already exist
-                Map<String, String> p = new URLConfigReader(url).call().getToAdd();
+                Map<String, Object> p = new URLConfigReader(url).call().getToAdd();
                 LOG.debug("Loaded : {}", url.toExternalForm());
                 props.putAll(p);
     
                 // Recursively load any files referenced by an @next property in the file
                 // Only one @next property is expected and the value may be a list of files
-                String next = p.get(INCLUDE_KEY);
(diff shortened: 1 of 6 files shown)


archaius2-core/src/main/java/com/netflix/archaius/readers/PropertiesConfigReader.java
```diff
@@ -64,16 +64,16 @@ private void internalLoad(Properties props, Set<String> seenUrls, ClassLoader lo
...
+                Map<String, Object> p = new URLConfigReader(url).call().getToAdd();
                 LOG.debug("Loaded : {}", url.toExternalForm());
                 props.putAll(p);
     
                 // Recursively load any files referenced by an @next property in the file
                 // Only one @next property is expected and the value may be a list of files
-                String next = p.get(INCLUDE_KEY);
-                if (next != null) {
+                Object next = p.get(INCLUDE_KEY);
+                if (next != null && next instanceof String) {
```
elandau: Next should only ever be a string.  It's OK to keep next as a String.

String next = (String)p.get(INCLUDE_KEY);

costimuraru: Thanks for the feedback, @elandau. Made the suggested change.

howardyuan:
Code LGTM. @costimuraru Is this part of your efforts to support java.time.Duration?

costimuraru:
Indeed it is @howardyuan.
One issue though: `java.time.Duration` comes with Java 8. I can see that archaius2 is built with Java 7. As you probably know, Java 7 has reached end of line as of April 2015 [0].
Do you see a problem with upgrading to Java 8? The alternative for me is to keep the changes on the fork, which is not very practical :-)

[0] https://java.com/en/download/faq/java_7.xml

PR #408, opened 2016-04-15
Visibility changes to support some extension

twicksell:
make MethodInvoker interface public
open visibility to methods on DefaultPropertyFactory for extension
expose Config object backing ConfigProxyFactory


elandau:
LGTM

PR #410, opened 2016-04-21
Fix isEmpty by isNotEmpty for overrided properties

gdespres:


rgallardo-netflix:

PR #412, opened 2016-05-04
1.x allow more than 1 dynamodb table as configuration source

hieurl:
Using them same property com.netflix.config.dynamo.tableName with delimiter ,
eg: com.netflix.config.dynamo.tableName=CommingProperties,MyAppProperties

Use case: allow multiple modules to share configuration while separating private properties.


howardyuan:
Code logic seems straightforward. I'm not familiar with Dynamo configuration source, is there a way that you can add some test cases for this?

hieurl:
Thanks for the comment @howardyuan 
I created the multiple table test class separately to avoid breaking the other tests since with multiple table, the number of mock scan request is different than single table

rgallardo-netflix:

PR #429, opened 2016-08-05
add handling for java 8 time classes and miscellaneous others

tcellucci:
period and duration are especially needed


gorzell:
Given that the source version of the build is set to 1.7 I am pretty sure this isn't going to compile without changing that.  I personally wouldn't mind that change, but it is pretty significant.

elandau:
LGTM

Issue #434, opened 2016-08-22
Get properties used by application


Aloren:
I would like to get all properties that are used by application.
For example in my configuration db I have:

```
property1=value1
property2=value2
property3=value3
property4=value4
```

and my application uses only `property1` and `property3` properties. I would like to have some kind of registry in which all actually used properties were held (in my case it would hold: `property1`,`property3`).
I have looked through the code, but haven't found something like that. As I understand the starting point for implementation of such feature is `DynamicProperty` class. Correct me if I am wrong. 
Thanks.

howardyuan:
Can you elaborate a bit more on your use case and context? Tracking property usage is indeed useful. However, it can be tricky. For one, many users don't use DynamicProperty to get the properties. Instead, they access the properties through configurations directly. If you plan to use the registry for further actions like removing properties, there can be consequences. Another thing I worry about is that now you're tagging everybody with some extra overhead even if they don't care for the usages. This is actually a big topic so we need to give it more thoughts.

Aloren:
> Can you elaborate a bit more on your use case and context?

In our application there are a lot of libraries that fetch settings in different ways. Sometimes it becomes really difficult to track which settings are actually used by application. 

> many users don't use DynamicProperty to get the properties. Instead, they access the properties through configurations directly.

Ok, I understand. Maybe there is another point where I can surely know that setting was requested by application?

> Another thing I worry about is that now you're tagging everybody with some extra overhead

This feature can be optional, by default it will be off.

howardyuan:
@Aloren Tracking the usage at per property level is a big change. At this point, we don't plan to implement it in Archaius 1. But we'll look into this in Archaius 2 and later versions.

PR #435, opened 2016-08-22
Issue 434: Get properties used by application

Aloren:
See [issue 434](https://github.com/Netflix/archaius/issues/434) for more details


howardyuan:
I think what you're trying to achieve here is already captured in ALL_PROPS static map in DynamicProperty. Right now there's no accessor. But adding a getter for it is a better solution. However, keep in mind that these two solutions won't capture those properties that aren't accessed at runtime. Different code paths can potentially yield different properties.

PR #437, opened 2016-08-30
Support configuration interface default method

elandau:
Support invoking the an interface method's default implementation when no configuration value is found.

For example,

```java
interface MyConfig {
   default int getTimeout() { 
       return 1000;
   }
}
```
Note that a method may only have a default implementation or a @DefaultValue annotation but not both. An IllegalArgumentException will be thrown if both are present.


twicksell:
Overall lgtm, but I feel like there is some javadoc or a wiki page that should be updated to point out the caveat of not using @DefaultValue and a default method. Does @Configuration describe these options?

PR #439, opened 2016-09-02
Basic support for Set, List types in proxy mapping

elandau:


archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -206,21 +215,29 @@ public T invoke(Object Obj, Object[] args) {
...
                 Object defaultValue = null;
                 if (m.getAnnotation(DefaultValue.class) != null) {
                     if (m.isDefault()) {
                         throw new IllegalArgumentException("@DefaultValue cannot be defined on a method with a default implementation for method "
                                 + m.getDeclaringClass().getName() + "#" + m.getName());
+                    } else if (
+                            Map.class.isAssignableFrom(returnType) ||
+                            List.class.isAssignableFrom(returnType) ||
+                            Set.class.isAssignableFrom(returnType) ) {
+                        throw new IllegalArgumentException("@DefaultValue cannot be used with collections.  Use default method implemenation instead "
```
tcellucci: why not?

elandau: To properly support this I'd need to make more changes than I'm comfortable making in this one PR.  I'd rather just support defaults via default methods and add support for @DefaultValue at a later time.

archaius2-api/src/main/java/com/netflix/archaius/api/PropertyContainer.java
```diff
@@ -85,4 +86,6 @@
      * should be optimized to call one of the known parsing methods based on type. 
      */
     <T> Property<T> asType(Class<T> type, T defaultValue);
+    
+    <T> Property<T> asType(Function<String, T> type, String defaultValue);
```
jhspaybar: Does this need a default implementation so it's backwards compatible?

archaius2-core/src/test/java/com/netflix/archaius/ProxyFactoryTest.java
```diff
@@ -231,4 +237,63 @@ public void testWithLongMap() {
...
+    
+    @Test
+    public void testCollectionsWithoutValue() {
+        SettableConfig config = new DefaultSettableConfig();
+        
+        PropertyFactory factory = DefaultPropertyFactory.from(config);
+        ConfigProxyFactory proxy = new ConfigProxyFactory(config, config.getDecoder(), factory);
+        ConfigWithCollections withCollections = proxy.newProxy(ConfigWithCollections.class);
+        
+        Assert.assertNull(withCollections.getLinkedList());
```
tcellucci: afaik best practice for apis that return collections is to return an empty list / map / set rather than null; can we do that here?

archaius2-core/src/test/java/com/netflix/archaius/ProxyFactoryTest.java
```diff
@@ -231,4 +237,63 @@ public void testWithLongMap() {
...
+        config.setProperty("sortedSet", "5,4,3,2,1");
+        config.setProperty("linkedList", "5,4,3,2,1");
+        
+        PropertyFactory factory = DefaultPropertyFactory.from(config);
+        ConfigProxyFactory proxy = new ConfigProxyFactory(config, config.getDecoder(), factory);
+        ConfigWithCollections withCollections = proxy.newProxy(ConfigWithCollections.class);
+        
+        Assert.assertEquals(Arrays.asList(5,4,3,2,1), new ArrayList<>(withCollections.getLinkedList()));
+        Assert.assertEquals(Arrays.asList(5,4,3,2,1), new ArrayList<>(withCollections.getList()));
+        Assert.assertEquals(Arrays.asList(5,4,3,2,1), new ArrayList<>(withCollections.getSet()));
```
jhspaybar: Can we assert order on an unsorted set?  I'd assume not.

elandau: I'm actually creating a LinkedHashSet so the order is preserved (not sorted).  Not sure it's necessary but felt like it would be nice to maintain.   I can change this to a HashSet if you feel that's more appropriate.

archaius2-api/src/main/java/com/netflix/archaius/api/PropertyContainer.java
elandau: Other than throwing an exception I'm not sure there's any valid default here.  I doubt anyone is implementing this anyway.

archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -293,6 +321,38 @@ public T invoke(Object Obj, Object[] args) {
...
+        };
+    }
+    
+    private MethodInvoker<?> createCollectionProperty(String propName, ParameterizedType type, Supplier<Collection> listSupplier) {
+        final Class<?> valueType = (Class<?>)type.getActualTypeArguments()[0];
+        return createCustomProperty(s -> { 
+            Collection list = listSupplier.get();
+            if (!s.isEmpty()) {
+                Arrays.asList(s.split("\\s*,\\s*")).forEach(v -> {
+                    if (!v.isEmpty() || valueType == String.class) {
```
twicksell: Strange that we should need this. Decoder should have a method supportsType to determine if the decoder implementation you're using supports the type.

elandau: Agreed.  A proper fix for this would require much more refactoring than I'm comfortable with with this PR.  Let's merge this as is and fix the Decoder and default value issue restrictions in a separate PR.

PR #444, opened 2016-09-15
DONT set archaius1 properties in ArchaiusModule

twicksell:
These properties should only be set when using StaticArchaiusBridgeModule


elandau:
LGTM

PR #445, opened 2016-09-21
Add closing quote to error msg

qiangdavidliu:


twicksell:
:thumbsup:

archaius2-core/src/main/java/com/netflix/archaius/DefaultDecoder.java
```diff
@@ -63,7 +63,7 @@
             else if (v.equalsIgnoreCase("false") || v.equalsIgnoreCase("no") || v.equalsIgnoreCase("off")) {
                 return Boolean.FALSE;
             }
-            throw new ParseException("Error parsing value '" + v, new Exception("Expected one of [true, yes, on, false, no, off]"));
+            throw new ParseException("Error parsing value '" + v + "'", new Exception("Expected one of [true, yes, on, false, no, off]"));
```
brharrington: Out of curiosity, why does it create a new exception with the expected values just to wrap it? Couldn't the expectation just be appended to the message for the ParseException?

qiangdavidliu: That's a fair question, I'll have to look into why that's the case.

PR #451, opened 2016-10-03
Record stack trace at time of ConfigurationManager creation

elandau:
When using the Archaius2 static bridge sometimes ConfigurationManager is initialized too early and it is currently not possible to identify the initialization path when the bridge fails to created.  The exposed initStack can now be added to StaticAbstractConfiguration so the static initialization path can be easily identified.


qiangdavidliu:
👍

PR #457, opened 2016-10-21
add java.net classes into decoder

tcellucci:


archaius2-core/src/main/java/com/netflix/archaius/DefaultDecoder.java
```diff
@@ -41,6 +44,7 @@
 import javax.inject.Singleton;
 import javax.xml.bind.DatatypeConverter;
 
+import com.google.common.net.HostAndPort;
```
qiangdavidliu: archaius2-core does not have a dependency on Guava

PR #459, opened 2016-10-25
Fix Typesafe exception handling.

gorzell:
Getting a property was throwing an exception, even in the case where a default value was provided.  This is not the correct behavior.  Also started adding some more tests for this kind of thing.


gorzell:
@elandau would be great to get this into the next release, it is a pretty significant bug.

gorzell:
I will do a separate PR for List support.

gorzell:
@elandau can I get a review and merge here?  My other PR too?

PR #460, opened 2016-10-26
Load multiple resources with the same resource name

elandau:
Currently property file loading will stop only at the first file is found.  This can be very problematic when the same property exists in multiple places on the classpath.  Since the file loading order is non deterministic its reasonable to simply load all found resources into a single CompositeConfig.  It is however possible for the same property to be defined in multiple files in which case the first loaded will win.


qiangdavidliu:
It would be good to have a test case for this, that also checks and ensures the override behaviour when the same prop is defined in multiple config files.

PR #461, opened 2016-10-26
@Provides method configuration loading

elandau:
Support configuration loading by annotating an @Provides method with @ConfigurationSource.

For example, to load foo.properties before getFoo() is called,

```java
new ArchaiusModule() {
    @Override
    protected void configureArchaius() {
    }

    @Provides
    @Singleton
    @ConfigurationSource("foo")
    Foo getFoo() {
          return new Foo();
    }
}
```


rgallardo-netflix:
We are no longer adding features specific to Guice

PR #465, opened 2016-11-16
Implement the Supplier interface

### LINKED ISSUE #464: Property<T> should implement Supplier<T>
A vast majority of our usage of a Property is calling .get() on it. To mock out that behavior we could mock out a DynamicStringProperty, for example, but it's much easier to define the type to be Supplier<T> and mock the Supplier interface. Supplier much better fits the most basic type we want to deal with in our code, hence we define the type as such. So currently we have to go through these hoops to make the value of a property testable:

```
public class DelayStartupHealthIndicator {
    private final Supplier<Long> delay;

    public DelayStartupHealthIndicator() {
        this(DynamicPropertyFactory.getInstance().getLongProperty("startupTimeDeylayInMinutes", 10));
    }

    private DelayStartupHealthIndicator(DynamicLongProperty startupDelay) {
        this(() -> startupDelay.get());
    }

   // Real constructor
   /* package */ DelayStartupHealthIndicator(Supplier<Long> startupDelayMinutes) {
        this.delay = delay;
    }
}
```

In the test, we can construct the object like this:

```
new DelayStartupHealthIndicator( () -> 10 )
```

That is much easier than trying to mock out DynamicLongProperty or DynamicPropertyFactory. It'd should be easy to implement the Supplier interface.

quidryan:
Fixes #464

### DIFF
--- a/archaius2-api/src/main/java/com/netflix/archaius/api/Property.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/api/Property.java
@@ -15,6 +15,8 @@
  */
 package com.netflix.archaius.api;
 
+import java.util.function.Supplier;
+
 /**
  * API to access latest cached value for a Property.  A Property is created from a PropertyFactory
  * that is normally bound to a top level configuration.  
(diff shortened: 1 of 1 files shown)


rgallardo-netflix:
Handled in #526

PR #466, opened 2016-11-30
Check for null node response before cacheing

TarzanInAJar:
init() does not handle the case where the Etcd object's list() command results in an error (maybe due to an empty configuration). A NullPointerException is thrown when cacheConfigValues tries to call "getNodes()" on a null pointer.

I made a change to init() that reports that an error has occurred/the response code if it happens, and doesn't attempt to cache anything.
I also made a change to cacheConfigValues that checks for null before calling getNodes() (which shouldn't happen now, but good practice to check anyway)


rgallardo-netflix:

PR #470, opened 2016-12-20
Update PropertyWrapper to be thread safe.

pkwarren:
Update the callbacks collection in PropertyWrapper to be thread safe
(similar to DynamicProperty). Remove runnable entries from callbacks
collection after removing from DynamicProperty.

### DIFF
--- a/archaius-core/src/main/java/com/netflix/config/PropertyWrapper.java
+++ b/archaius-core/src/main/java/com/netflix/config/PropertyWrapper.java
@@ -15,10 +15,11 @@
  */
 package com.netflix.config;
 
+import java.util.HashSet;
 import java.util.IdentityHashMap;
-import java.util.List;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 
-import com.google.common.collect.Lists;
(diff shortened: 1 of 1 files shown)


archaius-core/src/main/java/com/netflix/config/PropertyWrapper.java
```diff
@@ -158,9 +159,11 @@ public void addCallback(Runnable callback) {
      */
     @Override
     public void removeAllCallbacks() {
-        for (Runnable callback: callbackList) {
+        final Set<Runnable> callbacksToRemove = new HashSet<Runnable>(callbacks);
+        for (Runnable callback : callbacksToRemove) {
             prop.removeCallback(callback);
         }
+        callbacks.removeAll(callbacksToRemove);
```
howardyuan: Why not calling callbacks.clear() here?

pkwarren: Since other threads could add new callbacks while `removeAllCallbacks` is processing, this is just extra protection to prevent removing callbacks from the set that haven't had `removeCallback(Runnable)` called on them.

Issue #426, opened 2016-07-07
Archaius 2 should have its own pluggable logging


agentgt:
One of the reasons why my company has not switched over to Archaius 2 is that Archaius 2 core uses SLF4J (we have our own internal config library that [I need to finish open sourcing](https://github.com/agentgt/configfacade)).

The inherent problem of using SLF4J in a configuration library is that you now really can't use the library to configure any logging. The static binding nature of SLF4J further exacerbates this. Of course one solution would be to reinitialize the logging framework as a sort of two phase approach but this is not ideal particularly since even modern logging frameworks (e.g. logback) have issues at times with reinitialization. 

It would be ideal if Archaius 2 allowed a simple SPI for plugging in custom logging (perhaps through the Service Loader mechanism). I know this sort of defeats the purpose of SLF4J but this library is sort of a fringe case and I think configuration bootup should happen before logging framework initialization.

gorzell:
I am not sure I am understanding the issue here.  Do you have a use case where you would like to swap out the actual logging binding at runtime based on configuration information?  If so what does that look like?

Otherwise it seems like having logging in the core is pretty valuable.

agentgt:
Not the logging binding but the configuration. For example lets say you are publishing messages to Kafka or RabbitMQ you need the configuration of that server information. 

In some cases we use LogBack and some cases we use legacy Log4J. The most reliable way to configure both of these is through System Properties. Consequently when are app boots up it loads properties and will set some System Properties **before anything else gets loaded**. And this isn't just for logging as some legacy libraries use System Properties (this is because there isn't  a standard way to get configuration like there is for logging (e.g. SLF4J or java.util.logging).

SLF4J offers no way to intercept the binding. It is static and allows no hooks. I have ranted about this annoyance for some time (http://mailman.qos.ch/pipermail/logback-user/2016-June/004781.html).

Regardless even if SLF4J did support interception there are many high performance libraries and applications that prefer not do any initialization of logging *but still would like some way to do configuration* (otherwise we are back to System.properties).

agentgt:
Also Typesafe config does zero logging so clearly a configuration framework can be written with out needing logging.

PR #476, opened 2017-01-25
Add pluggable logging for #426 v2

agentgt:
See original issue. I'm of the high opinion that at least core configuration should not require logging otherwise you can't configure logging!!!

This change adds a private logging facade for archaius. I have set the default to NoOp but one could easily write a SLF4J and have it be the default. We could also use the ServiceLoader mechanism to load up a default logging implementation (I actually helped @mattrjacobs setup something similar for Hystrix plugins).

Of course the possibly simpler and perhaps better option is to not do any logging in `archaius2-core` and fail fast and configuration to ignore errors... etc.

Ignore the checked in pom file (I needed it because I'm mirroring in a corp environment with mercurial as a forked version and I can't edit the commit... yet). I will squash/edit the commit and remove it if there is interest in this addition.


agentgt:
I figured out a simpler way that requires less code if you want to get the dependency to SLF4j:

```java
package com.netflix.archaius.api.log;

import java.util.ServiceLoader;

import org.slf4j.ILoggerFactory;
import org.slf4j.Logger;
import org.slf4j.helpers.NOPLoggerFactory;

public interface LoggerService {
	
	public static class InternalLoggerFactory {

		/*
		 * Simply prefix Internal in front of LoggerFactory:
		 * So instead of:
		 * Logger logger = LoggerFactory.getLogger(MyClass.class);
		 * It should be:
		 * Logger logger = InternalLoggerFactory.getLogger(MyClass.class);
		 */
		public static Logger getLogger(Class<?> clazz) {
			return getILoggerFactory().getLogger(clazz.getName());
		}
		
		public static ILoggerFactory getILoggerFactory() {
			return LazyHolder.INSTANCE;
		}
		
	    //We should not load unless we are requested to. This avoids accidental initialization. @agentgt
	    //See https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom
		private static class LazyHolder { 
			private static final ILoggerFactory INSTANCE = InternalLoggerFactory.createFactory();
		}
		
		private static ILoggerFactory createFactory() {
			/*
			 * You could also use a system property or static variable for resolution as well.
			 * The following looks for a file located here:
			 *  /META-INF/services/LoggerService.getPackage().getName()
			 * With the name of the class as a single line that implements LoggerService.
			 */
			ServiceLoader<LoggerService> loader = ServiceLoader.load(LoggerService.class);
			for (LoggerService ls : loader) {
				return ls.createLoggerFactory();
			}
			return new NOPLoggerFactory();
		}

	}
	
	public ILoggerFactory createLoggerFactory();

}
```

PR #481, opened 2017-04-03
Fix wiki link in README

positron:
It won't render correctly with the space between the `]` and `(`.


rgallardo-netflix:

Issue #491, opened 2017-06-15
ConfigProxyFactory support @DefaultValue for Collections


quidryan:
Reference: https://github.com/Netflix/archaius/blob/2.x/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java#L229

ConfigProxyFactory will throw an exception if a @DefaultValue is is provided for a Collection. Having the DefaultValue be a String makes it operationally easier to copy and modify in live systems. While I appreciate the default implementation on the interface, @DefaultValue is a nice standardization.

CC @elandau

PR #492, opened 2017-06-15
Honor default methods for collections

elandau:
Fix default methods for collection types not being called for ConfigProxyFactory


twicksell:
#491

PR #493, opened 2017-06-15
Immutable state

elandau:
- Introduce a simpler PropertySource API
- Improve thread safety by using immutable datasets to track composite configurations.
- Introduce a functional paradigm to iterate properties thereby improving consistency and thread safety


archaius2-api/src/main/java/com/netflix/archaius/api/PropertySource.java
```diff
@@ -0,0 +1,31 @@
...
+     * Get the raw property value.  No interpolation or other modification is done to the property.
+     * @param key
+     */
+    default Optional<Object> getProperty(String key) { return Optional.empty(); }
+
+    /**
+     * Mechanism for consuming all properties of the PropertySource
+     * @param consumer
+     */
+    default void forEachProperty(BiConsumer<String, Object> consumer) {}
```
twicksell: Do we absolutely need this here? It doesn't match up with the Spring one, so just questioning if it really needs to be here or if its good enough on Config or some other sub-interface

elandau: Yes, we do.  This is used heavily in archaius to merge property sources and provides thread safety guarantees instead of having to externally lock getKeys() and call getProperty() when iterating a PropertySource.

PR #496, opened 2017-06-20
Build archaius-scala properly and support Scala 2.12

mmlac:
- Upgraded Gradle to 3.5 (can't build with <= 2.4)
- Updated Netflix Dependencies to latest version to support Gradle 3.5
- Added 2.11 & 2.12 build step with proper Scala naming of library
- Disabled JavaDoc linting tasks that is enabled by default on Java8
- Upgraded scalatest to 3.0.2 and adapted tests


mmlac-bv:
Pinging @gorzell @elandau. Any feedback or can you merge this? 
Scala 2.12 is out for 10 months now and archaius-scala doesn't work on it in it's current state.

smanikim:
@mfulgo do we have plans to merge this

mfulgo:
@smanikim - I'm not sure. Maybe try bringing it to the attention of a maintainer.

smanikim:
thanks @mfulgo 
@elandau @gorzell  do we have plans to merge this

Lexer11l:
Hi there,
We also use the library in our project. We started migration to scala 2.12 and get binary incompatibility.
@Landau @gorzell @qiangdavidliu is there any chance to get it merged and released? We would really appreciate that.

archaius-scala/src/main/scala/com/netflix/config/scala/ChainedBooleanProperty.scala
```diff
@@ -25,9 +25,12 @@ class ChainedBooleanProperty(
 extends ChainedProperty[Boolean]
 {
 
-  def this(prefix: Option[String], name: String, suffix: Option[String], default: Boolean, callback: Option[() => Unit] = None) = {
+  def this(prefix: Option[String], name: String, suffix: Option[String], default: Boolean, callback: Option[() => Unit]) = {
     this(ChainMakers.fanPropertyName(prefix, name, suffix), default, callback)
   }
+  def this(prefix: Option[String], name: String, suffix: Option[String], default: Boolean) = {
```
gorzell: **Question:** Why all of the additional signatures here? It's been a long while since I have written any Scala, did something change in more recent versions that makes this necessary?

.gitignore
```diff
@@ -47,6 +47,9 @@ Thumbs.db
 */target
 /build
 */build
+*/build_2.10
```
gorzell: **Nit:** I think that at least for the moment this is extraneous?

mmlac:
@gorzell Thanks for the review. I honestly don't remember exactly why I made these changes 2.5 years ago. The changes to the .gitignore are because of cross-builds. 2.10 and 2.11 can probably be skipped from building today entirely.
For the signature, I think I just wanted to eliminate the double defaulting of the callback, once in the definition of the class and then in the overridden this, so I just made it two different constructors. Purely stylistic change if I remember correctly and can be ignored if desired.

rgallardo-netflix:
Closed because 1.x is no longer maintained and 2.x does not have scala bindings.

PR #512, opened 2017-08-29
New LayeredConfig API

elandau:
LayeredConfig will eventually replace CompositeConfig as the top level container for properties.  LayeredConfig simplifies defining multiple override layers that may be added in any order while keeping a well-defined override order.  A set of common layers is defined in the Layers class but additional user defined layers may be defined and added at any time.

This PR is the first in the series of changes through which the LayeredConfig will eventually be the top level config API while keeping backwards compatibility in mind.  Once dependent libraries support a code path for LayeredConfig as the top level API, ArchaiusModule will be updated to use LayeredConfig instead of CompositeConfig.

### DIFF
--- a/archaius2-api/src/main/java/com/netflix/archaius/Layers.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/Layers.java
@@ -1,0 +1,18 @@
+package com.netflix.archaius;
+
+import com.netflix.archaius.api.Layer;
+
+public final class Layers {
+    public static Layer TEST        = Layer.of("test",        100, false);
+    public static Layer RUNTIME     = Layer.of("runtime",     200, false);
+    public static Layer SYSTEM      = Layer.of("system",      300, false);
+    public static Layer ENVIRONMENT = Layer.of("environment", 400, false);
+    public static Layer REMOTE      = Layer.of("remote",      500, false);
+    public static Layer APPLICATION_OVERRIDE = Layer.of("application-override", 600, true);
+    public static Layer APPLICATION = Layer.of("application", 700, true);
+    public static Layer LIBRARY     = Layer.of("library",     800, false);
+    public static Layer DEFAULT     = Layer.of("default",     900, false);
+    
+    private Layers() {
+    }
+}
--- a/archaius2-api/src/main/java/com/netflix/archaius/Layers.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/Layers.java
@@ -3,15 +3,15 @@ package com.netflix.archaius;
 import com.netflix.archaius.api.Layer;
 
 public final class Layers {
-    public static Layer TEST        = Layer.of("test",        100, false);
-    public static Layer RUNTIME     = Layer.of("runtime",     200, false);
-    public static Layer SYSTEM      = Layer.of("system",      300, false);
-    public static Layer ENVIRONMENT = Layer.of("environment", 400, false);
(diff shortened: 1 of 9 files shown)


archaius2-api/src/main/java/com/netflix/archaius/Layers.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.netflix.archaius.api.Layer;
+
+public final class Layers {
+    public static Layer TEST        = Layer.of("test",        100, false);
+    public static Layer RUNTIME     = Layer.of("runtime",     200, false);
+    public static Layer SYSTEM      = Layer.of("system",      300, false);
+    public static Layer ENVIRONMENT = Layer.of("environment", 400, false);
+    public static Layer REMOTE      = Layer.of("remote",      500, false);
+    public static Layer APPLICATION_OVERRIDE = Layer.of("application-override", 600, true);
```
twicksell: Doesn't application-override typically beat remote?

archaius2-api/src/main/java/com/netflix/archaius/api/Layer.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Key used to group and order {@link Config}s into layers (@see Layers).
+ * Layers are ordered by natural order such that lower order values have 
+ * precedence over a higher value.  Within a layer configurations are prioritized by
+ * insertion order (or reversed if 'reversed=true')
+ */
+public final class Layer {
+    private final String name;
+    private final int order;
+    private final boolean reversed;
```
twicksell: Why do we need to keep this concept around? Is this truly necessary for a new API?

archaius2-api/src/main/java/com/netflix/archaius/api/config/LayeredConfig.java
```diff
@@ -0,0 +1,47 @@
...
+import com.netflix.archaius.api.Layer;
+import com.netflix.archaius.api.exceptions.ConfigException;
+
+import java.util.Collection;
+import java.util.Optional;
+
+/**
+ * Composite Config where the override order is driven by Layer keys.
+ */
+public interface LayeredConfig extends Config {
```
twicksell: Long-term what happens to the Config interface? Would we even want to keep it? Will we continue to expose bindings for it? (beyond whatever is needed for backwards compatibility)

archaius2-api/src/main/java/com/netflix/archaius/api/Layer.java
elandau: I suppose we don't need this and can add it later if the need should arise.

archaius2-api/src/main/java/com/netflix/archaius/api/config/LayeredConfig.java
elandau: Yes, this is mostly for backwards compatibility.  Long term I'd like to get rid of Config in favor of a PropertyResolver and PropertySource, similar to what Spring has.  This should ease the migration in the future as well.  Once this has been released I'll be making more changes towards introducing PropertyResolver and PropertySource derived interfaces.

archaius2-api/src/main/java/com/netflix/archaius/Layers.java
twicksell: Nevermind, mixed up runtime and application-override in my head.

Issue #486, opened 2017-05-08
AbstractConfig uses exceptions for control flow


brharrington:
The `getValueWithDefault` method in AbstractConfig is using an exception to trigger using the default value:

https://github.com/Netflix/archaius/blob/2.x/archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java#L250

```java
    protected <T> T getValueWithDefault(Class<T> type, String key, T defaultValue) {
        try {
            return getValue(type, key);
        } catch (NoSuchElementException e) {
            return defaultValue;
        }
    }
```

This add considerable overhead to the methods that use a default.

/cc @dmuino @drather19

brharrington:
Fixed in #523

PR #521, opened 2017-10-15
Use an exception with no stack as flow control for getDefaultValue

jhspaybar:


jhspaybar:
This should address https://github.com/Netflix/archaius/issues/486 .  We still use exceptions for flow control, but just don't make the stack which is the expensive part.

elandau:
How about we just deprecate all the methods where a default value is not provided.  That path does not throw an exception if the key is not found.

jhspaybar:
@elandau the issue is this block ```protected <T> T getValueWithDefault(Class<T> type, String key, T defaultValue) {
        try {
            return getValue(type, key);
        } catch (NoSuchElementException e) {
            return defaultValue;
        }
    }```
it doesn't bubble an exception outside of archaius, but the throw/catch here to use the default value is actually what causes the excess CPU usage.

elandau:
Ah, I was just looking at the getString() case.  In that case let's flip the implementation between getValue and getValueWithDefault so that getValue() throws an exception if the value is null.

PR #524, opened 2017-10-30
add property-repo subproject

tcellucci:
PropertyRepo = type-neutral property interface with bindings to Archaius 1, Archaius 2 or other property systems


elandau:
Replaced with #526

PR #530, opened 2017-11-14
API to unsubscribe to property change notifications

elandau:
Sample usage

```
Property<Integer> prop = registry.get("foo", Integer.class);
Subscription subscription = prop.onChange((newValue) -> System.out.println("new value: " + newValue));
```
To unsubscribe

````
subscription.unsubscribe();
```

### DIFF
--- a/archaius2-api/src/main/java/com/netflix/archaius/api/Property.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/api/Property.java
@@ -44,6 +44,14 @@ import java.util.function.Supplier;
  */
 public interface Property<T> extends Supplier<T> {
     /**
+     * Token returned when calling onChange through which change notification can be
+     * unsubscribed.  
+     */
+    interface Subscription {
+        void unsubscribe();
+    }
+    
+    /**
      * Return the most recent value of the property.  
      * 
      * @return  Most recent value for the property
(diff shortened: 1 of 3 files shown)


archaius2-api/src/main/java/com/netflix/archaius/api/Property.java
```diff
@@ -77,16 +85,20 @@ default void removeListener(PropertyListener<T> listener) {}
...
      * up until this point. Changes to subsequent Property objects returned from {@link Property#orElse} 
      * or {@link Property#map(Function)} will not trigger calls to this consumer.
+
      * @param consumer
-     * @return This property object
+     * @return Subscription that may be unsubscribed to no longer get change notifications
      */
-    default void onChange(Consumer<T> consumer) {
-        addListener(new PropertyListener<T>() {
+    default Subscription onChange(Consumer<T> consumer) {
```
tcellucci: nit: semantically 'Subscription' isn't a very obvious result from the 'onChange' method; would 'subscribe' work as the method name here?  What is the terminology in RexJava?

archaius2-api/src/main/java/com/netflix/archaius/api/Property.java
```diff
@@ -44,6 +44,14 @@
  */
 public interface Property<T> extends Supplier<T> {
     /**
+     * Token returned when calling onChange through which change notification can be
+     * unsubscribed.  
+     */
+    interface Subscription {
```
tcellucci: doesn't this need to be explicitly public for appropriate visibility?

elandau: nope

PR #536, opened 2017-12-22
fix inconsistent use of prefix in PrefixedViewConfig

tcellucci:
PrefixedViewConfig had inconsistent handling of the 'prefix' constructor arg; sometimes it used the raw value (passed into State) and sometimes it used the normalized value (modified to always end with a '.').  

Previously if the prefix constructor arg for PrefixedViewConfig ended with a '.', then property values returned via PrefixedViewConfig would be incorrect; too short by one character.


qiangdavidliu:
👍

rgallardo-netflix:
@akang31 , is this PR adressing the same thing you fixed in #635 ?

akang31:
Yes, this is addressing the same issue which is now fixed.

PR #541, opened 2018-02-06
ConfigProxyFactory proxies should implement equals and hashCode

twicksell:


archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -230,8 +230,13 @@ public T invoke(Object[] args) {
             if (invoker != null) {
                 return invoker.invoke(args);
             }
-            
-            if ("toString".equals(method.getName())) {
+            if ("equals".equals(method.getName())) {
```
tcellucci: this will work ok for call low-volume, but a linear search for the methods by names isn't the greatest.  It'd be cleaner to put the Object.equals, .hashCode and .toString Methods into the invokers map with the others and make a lambda implementation of the internal MethodInvoker interface for each Object method

twicksell: We need data from InvocationHandler which isn't passed down through the MethodInvoker interface in that map. Doesn't really seem worth refactoring all of that.

PR #544, opened 2018-03-28
add timeout for URLConfigurationSource when retrieving configurations

ylgrgyq:
Hi, thanks for this great lib.

For this PR, I think we need to add a basic timeout when retrieving configurations in URLConfigurationSource. Because `openStream` method of `java.net.URL` will open a connection to the target URL without connection timeout or read timeout. If the target website or the connection have some issue, the thread using to poll configurations dynamically may be blocked forever and there's no any error log to let people notice this.

The hardcoding timeout value may seems inappropriate but I think it's OK because mostly archaius will be used in a stable environment with good network so this timeout is just a safety net and we don't even need to adjust it.

### DIFF
--- a/archaius-core/src/main/java/com/netflix/config/sources/URLConfigurationSource.java
+++ b/archaius-core/src/main/java/com/netflix/config/sources/URLConfigurationSource.java
@@ -18,6 +18,7 @@ package com.netflix.config.sources;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.net.URLConnection;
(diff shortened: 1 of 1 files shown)


juvenn:
Hi @allenxwang, any chance to take a look at this issue? We do have found configuration values not propagating to some of JVM instances due to thread hanging up here.

Would greatly appreciate if a release could be cut with this fix, thanks!

liuchong (approved):
👍

rgallardo-netflix:

PR #564, opened 2018-11-09
Introduce static Property implementation

akhaku:
A StaticProperty is useful for cases where we don't want a
property to change, for example in unit tests.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/StaticProperty.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/StaticProperty.java
@@ -1,0 +1,43 @@
+/**
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.archaius;
+
+import com.netflix.archaius.api.Property;
+
+/**
+ * A static {@link Property} whose value never changes. This is useful when creating
+ * properties for use in tests.
+ */
+public class StaticProperty<T> implements Property<T> {
+
+    private final String key;
+    private final T value;
+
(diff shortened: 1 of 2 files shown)


akhaku:
@twicksell - we discussed adding a method to the Property interface, but I'm not too sure how I feel about adding a factory (default) method to an interface, so I opted to stick with the separate implementation.

elandau:
Can you provide additional information about the use case here.  Property is normally created by a PropertyRepository and not directly so it feels strange to be creating a static property instance outside of a PropertyRepository.

akhaku:
Absolutely, so a couple different use cases:
1. In tests, to test methods that use a Property, it's helpful to pass in a static property instead of mocking or having an overloaded method.
2. In some non-DI cases (eg static instantiation for local development), we don't have Guice/other DI and so don't have a way of fetching a PropertyRepository. For those cases, it makes sense to have a static property to get the code working, albeit with default values for properties.

archaius2-core/src/main/java/com/netflix/archaius/StaticProperty.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package com.netflix.archaius;
+
+import com.netflix.archaius.api.Property;
+
+/**
+ * A static {@link Property} whose value never changes. This is useful when creating
+ * properties for use in tests.
+ */
+public class StaticProperty<T> implements Property<T> {
```
brharrington: Minor suggestion, consider calling it something other than static (immutable perhaps). Given the widespread use of static variables with archaius1, my first thought when seeing the PR title was completely different from what is implemented here. I think a different name would help avoid confusion when discussing it.

elandau:
I'd be concerned about introducing this given that archaius is all about dynamic configuration and I don't feel that this is the proper abstraction for introducing 'immutable' configuration.  I'd also recommend not relying on the Property interface directly in your code to avoid coupling your business logic specifically with archaius.  There are several alterantives to achieving this,
1.  Property implements Java's Supplier so you can just code to Supplier.  For testing you can easily create an immutable supplier like this, `() -> CONSTANT`.
2.  Define a configuration interface that is bound to archaius in the normal case or manually constructed for testing. 
3.  Create an immutable Config for your tests and bind that to a PropertyRepository or a configuration interface as a means to achieve immutable configuration.

akhaku:
Hm, interesting - we've done something similar in our code, where we have an interface that is implemented using an Archaius PropertyRepository. For non-DI (static creation for local development) we don't pass in a RepositoryFactory and instead initialize properties to static values, but we could instead use a DefaultPropertyFactory and have that use a SettableConfig. I'll give that a shot and add you to the PR so you can take a look, if you're curious.

PR #577, opened 2019-07-02
Fix issue with config proxies and Map values

kilink:
Because the ConfigProxyFactory evaluates the default value eagerly, an NPE is thrown for
Map properties when a null value is passed into Collections.unmodifiableMap, even if the
configuration is actually present and therefore the default is not needed.


howardyuan:
lgtm

rgallardo-netflix:
I believe this issue does not apply to the current implementation.

PR #578, opened 2019-08-15
Improved support for collection types

elandau:
- Introduce a new TypeConverter contract through which the decoder may be extended.
- Capture array, enum and typed collection conversions via TypeConverter.Factory implementations
- Support typed collections, such as `Map<String, Long>`, when using ConfigProxyFactory without having to wrap them in a custom type

### DIFF
--- a/archaius2-api/src/main/java/com/netflix/archaius/api/Decoder.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/api/Decoder.java
@@ -27,7 +27,11 @@ public interface Decoder {
 	<T> T decode(Class<T> type, String encoded);
 
 	default <T> T decode(Type type, String value) {
-		return decode((Class<T>)type, value);
+		if (type instanceof Class) {
+			return decode((Class<T>)type, value);
+		} else {
+			throw new UnsupportedOperationException("This decoder " + getClass() + " does not support Type");
+		}
 	}
 
 }
--- a/archaius2-api/src/main/java/com/netflix/archaius/api/Decoder.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/api/Decoder.java
@@ -15,11 +15,19 @@
  */
 package com.netflix.archaius.api;
 
+import java.lang.reflect.Type;
+
 /**
  * API for decoding properties to arbitrary types.
  *
  * @author spencergibb
  */
 public interface Decoder {
+	@Deprecated
 	<T> T decode(Class<T> type, String encoded);
+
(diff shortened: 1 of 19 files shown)


archaius2-core/src/main/java/com/netflix/archaius/converters/ArrayTypeConverterFactory.java
```diff
@@ -0,0 +1,37 @@
...
+            TypeConverter elementConverter = registry.get(clsType.getComponentType()).orElseThrow(() -> new RuntimeException());
+            return Optional.of(create(elementConverter, clsType.getComponentType()));
+        }
+
+        return Optional.empty();
+    }
+
+    private static TypeConverter<?> create(TypeConverter elementConverter, Class type) {
+        return value -> {
+            String[] elements = value.split(",");
```
howardyuan: any possibility that delimiter is something else? also, I'm wary of possible empty or null values. Do we protect the converter from such scenarios?

howardyuan:
I don't see new tests for enum/array collections since they're newly supported. I'm not sure if existing tests cover all the parameter typed collections without custom wrappers? If not, should we add tests as well? In general, I'm a bit wary of input checking in some places. My biggest fear is a mal-formed value causing runtime exception being thrown and apps won't be able to start.

archaius2-api/src/main/java/com/netflix/archaius/api/Decoder.java
```diff
@@ -15,11 +15,19 @@
...
  * API for decoding properties to arbitrary types.
  *
  * @author spencergibb
  */
 public interface Decoder {
+	@Deprecated
 	<T> T decode(Class<T> type, String encoded);
+
+	default <T> T decode(Type type, String value) {
+		return decode((Class<T>)type, value);
```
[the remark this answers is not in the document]
elandau: This is really here just so we don't break the contract now that we've added this method.  I'll update the default implementation to throw a more meaningful error when decoding something that is not a Class.

archaius2-core/src/main/java/com/netflix/archaius/converters/ArrayTypeConverterFactory.java
elandau: Code here is copied from the existing version so there is no behavior change.  When converting to an array I think it's fine to fail for null or missing values.

howardyuan:
I double checked current Archaius2 reader and Polling reader just treat property values as strings. So current changes are safe for properties loading. LGTM!

PR #581, opened 2019-09-03
Dynamic loading notification

elandau:
We need more complete notifications for configuration reloading, specifically we need an event before and after loading to control fast property behavior.


archaius-core/src/main/java/com/netflix/config/AbstractPollingScheduler.java
```diff
@@ -160,19 +162,20 @@ public void run() {
                 log.debug("Polling started");
                 PollResult result = null;
                 try {
+                    fireEvent(EventType.POLL_BEGIN, null, null);
                     result = source.poll(false, getNextCheckPoint(checkPoint));
                     checkPoint = result.getCheckPoint();
+
+                    try {
+                        populateProperties(result, config);
+                    } catch (Exception e) {
```
josephharrington: Why change the catches from `Throwable` to `Exception` here, but not above?

PR #610, opened 2020-02-20
Update DefaultPropertyFactory.java

smadappa:
A more descriptive output for the Property and its value


rgallardo-netflix:
Equivalent code has been merged already

PR #613, opened 2020-06-08
DynamicMapProperty escaping equal sign

eomiks:
DynamicMapProperty extracts key and value from keyValue string splitted by '='.
If value contains '=' char, keyValue string has more than two '='s, therefore after splitting DynamicMapProperty can't extract key&value, then just ignores it.
so, user can't set values that contains '=' char. and user get no error or warning, Key&values gonna just missing.
 
archaius should provide proper excaping.

I made a change to ```=``` can be escaped by ```\``` (in java ```"\\="```)


rgallardo-netflix:

PR #617, opened 2021-02-26
Add public modifiers to relevant methods.

### LINKED ISSUE #616: Allow Configuration Source Updates
Due to the issues raised in [this](https://github.com/Netflix/archaius/issues/46) discussion, it makes sense to allow the underlying property source to be modified or refreshed after it has already been initialized. Not allowing modification in the underlying source results in convoluted test setups, which can add both [complexity and unnecessary execution time](https://github.com/Netflix/archaius/issues/46#issuecomment-28715958) when requiring each test to run in a separate JVM in order to be able to change the underlying config source. 

Simply giving `public` access to the `DynamicProperty.updateProperty` and `DynamicProperty.updateAllProperties` here should be sufficient.

mpaauw:
Resolves #616.

### DIFF
--- a/archaius-core/src/main/java/com/netflix/config/DynamicProperty.java
+++ b/archaius-core/src/main/java/com/netflix/config/DynamicProperty.java
@@ -585,7 +585,7 @@ public class DynamicProperty {
     }
 
     // return true iff the value actually changed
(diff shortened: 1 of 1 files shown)


rgallardo-netflix:
V1 is no longer maintained.

PR #620, opened 2021-03-22
Filter out empty file name in the config urls

liuchong:
Filter out empty file name in the config urls.
This could also be solved with one line like https://github.com/leancloud/archaius/pull/2 but at least java8 is required:
``` java
fileNames = Arrays.stream(fileNames).filter(s -> !s.isEmpty()).toArray(String[]::new);
```


rgallardo-netflix:

liuchong:
Dose v2 fixed this? @rgallardo-netflix

rgallardo-netflix:
No, the equivalent code in v2 would just throw an exception for a null or empty name. Will track as new issue #644, because the code is too different to attempt just rebasing this PR.

PR #621, opened 2021-05-27
Correctly store lists in MapConfig

akhaku:


akhaku:
After some discussion internally we decided to leave this as-is.

akhaku:
Summary of the internal discussion:
- do we need to worry about commas in values? perhaps
- what if we instead changed it to Map<String, Object>?
- sounds good but that might not be compatible
- also unclear since java property values are always strings
- `MapConfig.Builder#put(String, T)` is misleading
- there's a workaround in app code so may not be worth making any changes here, at least until someone else compains

PR #624, opened 2021-09-03
Provide a proper error message instead of a ClassCastException

nikitar:
Otherwise `(String)newValue` fails and there's no way to know which property it was for.

There's currently one failing test, though for me it is failing on a clean branch as well.


rgallardo-netflix:

PR #628, opened 2022-08-16
Fixed interaction of default methods and parametrized properties.

rgallardo-netflix:
Parametrized properties with a default method fail because the code tries to create a MethodHandle to a method with no arguments. And, even if the method lookup is fixed, getting the default value for the property fails because the arguments to the call are not propagated.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
@@ -30,7 +30,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
-import java.util.function.Supplier;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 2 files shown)


archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -368,17 +356,26 @@ protected <T> MethodInvoker<T> createInterfaceProperty(String propName, final T
         return (args) -> proxy;
     }
 
-    protected <T> MethodInvoker<T> createScalarProperty(final Type type, final String propName, Supplier<T> next) {
+    protected <T> MethodInvoker<T> createScalarProperty(final Type type, final String propName, Function<Object[], T> next) {
```
asibross, line 359: This is an API change. But I doubt anyone is actually overriding these methods.
I looked internally and couldn't any reference.

DanielThomas (approved):
Let me know what release this goes out in and I'll update the JDK 17 upgrade guide w/ the details.

rgallardo-netflix:
@DanielThomas, this got released as 2.3.17

PR #629, opened 2022-10-28
Avoid Optional and lambda allocation in MethodInvoker

kilink:
Avoid allocating an Optional and lmabda on every call to the MethodInvoker returned by createScalarProperty.


rgallardo-netflix (approved):
Thanks

PR #634, opened 2023-02-03
Fix memory leak caused by anonymous listeners in abandoned views

akang31:
Views created that depend on other views (Composite, Layered, and Prefixed) create listeners on the source view that update themselves. This reference to themselves prevents GC until the source view is abandoned, meaning creating unused dependent views is a memory and performance leak. This applies the WeakReference logic from the PrivateView to the other dependent view types in order to fix this leak.

There is a small bug in PrefixedViewConfig where we supposedly allow prefixes of form "foo" or "foo.", but any prefix of form "foo." will actually still assume form "foo", meaning that configs such as "foo.bar" would, in the prefixed view, be taken as configs of type "ar" instead of the expected "bar".  I'll address this in a followup change in case there is anyone relying on this behavior.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/config/DependentConfigListener.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/config/DependentConfigListener.java
@@ -14,7 +14,7 @@ import java.util.Optional;
  *
  * @param <T> The type of the dependent config
  */
-public abstract class DependentConfigListener<T extends AbstractConfig> implements ConfigListener {
+abstract class DependentConfigListener<T extends AbstractConfig> implements ConfigListener {
     private final Reference<T> dependentConfigRef;
 
     DependentConfigListener(T dependentConfig) {
@@ -23,22 +23,22 @@ public abstract class DependentConfigListener<T extends AbstractConfig> implemen
 
     @Override
     public void onConfigAdded(Config config) {
-        updateState(config).ifPresent(vc -> onSourceConfigAdded(vc, config));
+        updateState(config).ifPresent(depConfig -> onSourceConfigAdded(depConfig, config));
     }
 
     @Override
     public void onConfigRemoved(Config config) {
-        updateState(config).ifPresent(vc -> onSourceConfigRemoved(vc, config));
+        updateState(config).ifPresent(depConfig -> onSourceConfigRemoved(depConfig, config));
     }
 
     @Override
     public void onConfigUpdated(Config config) {
-        updateState(config).ifPresent(vc -> onSourceConfigUpdated(vc, config));
+        updateState(config).ifPresent(depConfig -> onSourceConfigUpdated(depConfig, config));
     }
 
     @Override
     public void onError(Throwable error, Config config) {
-        updateState(config).ifPresent(vc -> onSourceError(error, vc, config));
+        updateState(config).ifPresent(depConfig -> onSourceError(error, depConfig, config));
     }
 
     public abstract void onSourceConfigAdded(T dependentConfig, Config sourceConfig);
--- a/archaius2-core/src/main/java/com/netflix/archaius/config/DependentConfigListener.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/config/DependentConfigListener.java
@@ -10,7 +10,7 @@ import java.util.Optional;
 /**
  * ConfigListener for the dependent/wrapper config paradigm. Most notably makes the reference to the dependent config
  * a weak reference and removes this listener from the source config so that any abandoned dependent configs can be
- * properly garbage collected
+ * properly garbage collected.
  *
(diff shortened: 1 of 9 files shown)


akang31:
Updated the PR to remove the fixes to PrefixedViewConfig. I'll address that in a followup PR.

archaius2-core/src/main/java/com/netflix/archaius/config/DependentConfigListener.java
```diff
@@ -0,0 +1,65 @@
...
+public abstract class DependentConfigListener<T extends AbstractConfig> implements ConfigListener {
+    private final Reference<T> dependentConfigRef;
+
+    DependentConfigListener(T dependentConfig) {
+        dependentConfigRef = new WeakReference<>(dependentConfig);
+    }
+
+    @Override
+    public void onConfigAdded(Config config) {
+        updateState(config).ifPresent(vc -> onSourceConfigAdded(vc, config));
```
rgallardo-netflix, line 26: Let's use more descriptive names here (and in the other methods too, of course)
I'd say something like
`.ifPresent(depConfig -> onSourceConfigAdded(depConfig, srcConfig));`

archaius2-core/src/main/java/com/netflix/archaius/config/PrefixedViewConfig.java
```diff
@@ -52,33 +52,46 @@ public State(Config config, String prefix) {
...
+        }
+
+        @Override
+        public void onSourceError(Throwable error, PrefixedViewConfig pvc, Config config) {
+        }
+    }
     
     public PrefixedViewConfig(final String prefix, final Config config) {
         this.config = config;
-        this.prefix = prefix.endsWith(".") ? prefix : prefix + ".";
```
rgallardo-netflix, line 58: I'm confused here. Did you or did you not revert the bug fix?

archaius2-core/src/test/java/com/netflix/archaius/config/CompositeConfigTest.java
```diff
@@ -146,6 +149,19 @@ public void getKeysTest() throws ConfigException {
         Assert.assertEquals(set("b1", "b2", "d1", "d2", "e1", "e2"), set(iter));
     }
 
+    @Test
+    public void unusedCompositeConfigIsGarbageCollected() throws ConfigException {
+        com.netflix.archaius.api.config.CompositeConfig config = DefaultCompositeConfig.builder()
```
rgallardo-netflix, line 154: Did you try running this code *before* the fix? I'm not sure the bug would be triggered in this case.

archaius2-core/src/main/java/com/netflix/archaius/config/DependentConfigListener.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * ConfigListener for the dependent/wrapper config paradigm. Most notably makes the reference to the dependent config
+ * a weak reference and removes this listener from the source config so that any abandoned dependent configs can be
+ * properly garbage collected.
+ *
+ * @param <T> The type of the dependent config
+ */
+public abstract class DependentConfigListener<T extends AbstractConfig> implements ConfigListener {
+    private final Reference<T> dependentConfigRef;
```
rgallardo-netflix, line 18: Let's make this class package-private, it's really just an implementation detail for the other config classes.

archaius2-core/src/main/java/com/netflix/archaius/config/DependentConfigListener.java
akang31: Ah good call. Updated to depConfig.

archaius2-core/src/main/java/com/netflix/archaius/config/PrefixedViewConfig.java
akang31: The previous code used "prefix" from the args instead of "this.prefix" in the anonymous class, meaning that "this.prefix" was never used (the version with the trailing dot added). In order to keep the behavior the same as before, we remove this logic (which is why we still have the prefix.length()+1) since now we actually start using this.prefix.

archaius2-core/src/test/java/com/netflix/archaius/config/CompositeConfigTest.java
akang31: Oops, I originally had a more verbose version of these tests and I forgot to keep around the reference to the source config while trimming them down. Re-added the reference to the source config and confirmed that these tests now fail against the previous version of the code.

archaius2-core/src/main/java/com/netflix/archaius/config/DependentConfigListener.java
akang31: Done

PR #635, opened 2023-02-09
Update PrefixedViewConfig to properly handle trailing dot case

akang31:
Current behavior for PrefixedViewConfig is that, for a prefix of form "foo." we expect configs of form "foo..bar", which is unusual.

We update PrefixedViewConfig to properly accept prefixes of form "foo." and expect configs of form "foo.bar".

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/config/PrefixedViewConfig.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/config/PrefixedViewConfig.java
@@ -47,7 +47,7 @@ public class PrefixedViewConfig extends AbstractConfig {
             data = new LinkedHashMap<String, Object>();
             config.forEachProperty((k, v) -> {
                 if (k.startsWith(prefix)) {
-                    data.put(k.substring(prefix.length()+1), v);
+                    data.put(k.substring(prefix.length()), v);
                 }
             });
         }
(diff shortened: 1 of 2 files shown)


rgallardo-netflix (commented):
Two requests:
1. Please go through the open issues here and see if anyone has reported this before.
2. Try to trace through the equivalent classes in Archaius 1. I'm curious if this is a regression from there, or if this weird behavior is long-standing.

akang31:
1. I went through the issues and I don't see any reports that match this specific bug.
2. I couldn't find an equivalent in Archaius 1. AFAICT, the PrefixedViewConfig is something that is specific to Archaius2. The specific change seems to have been made ~6 years ago in this change: https://github.com/Netflix/archaius/commit/6c3acc1584369c6847c83d1576ffed9431714c7f#diff-10f5c96baf2408d403c639812b33ccd7830da34e37f2a59674c716cef813569a (while changing the lambdas, the +1 was added to cover the bug from using prefix instead of this.prefix). The previous version of the file before that (https://github.com/Netflix/archaius/blame/161f437af8ab0242048bc91c97108d29395687ee/archaius2-core/src/main/java/com/netflix/archaius/config/PrefixedViewConfig.java) worked like it would following this change.

rgallardo-netflix:
👍 
Ah, got it.
PR LGTM :-)

rgallardo-netflix (approved):

akang31:
Thanks for the review!

PR #646, opened 2023-04-07
Revert Jackson dep to 2.4.3

akang31:
The update to 2.14.2 is causing issues in some downstream dependency resolution. We did two updates, one from 2.4.3 to 2.10.3 and then again to 2.14.2, but since we didn't have a version release in between, I'm reverting to the 2.4.3 for safety.


rpalcolea (approved):
LGTM

I cloned and ran dependencies with a scan to verify -> https://scans.gradle.com/s/6ggr76x6ove66/dependencies?dependencies=jackson&expandAll

PR #649, opened 2023-05-11
Add ability to get an Iterable of keys from a Config

kilink:
The existing `getKeys()` method of Config is problematic in practice in common usage, since it returns an `Iterator` instead of `Iterable`, which means callers have no ability to determine the size of the returned set of keys without fully iterating over it.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/config/DefaultSettableConfig.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/config/DefaultSettableConfig.java
@@ -36,23 +36,23 @@ public class DefaultSettableConfig extends AbstractConfig implements SettableCon
     public DefaultSettableConfig() {
         super(generateUniqueName("settable-"));
     }
-    
+
     @Override
     public synchronized <T> void setProperty(String propName, T propValue) {
         Map<String, Object> copy = new HashMap<>(props.size() + 1);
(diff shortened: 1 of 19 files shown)


archaius2-core/src/main/java/com/netflix/archaius/config/DefaultSettableConfig.java
```diff
@@ -36,16 +36,16 @@ public DefaultSettableConfig(String name) {
...
     }
-    
+
     @Override
     public synchronized <T> void setProperty(String propName, T propValue) {
         Map<String, Object> copy = new HashMap<>(props.size() + 1);
         copy.putAll(props);
         copy.put(propName, propValue);
-        props = copy;
+        props = Collections.unmodifiableMap(copy);
```
rgallardo-netflix, line 45: Please update all the other methods that alter the internal state to do this, too.

rgallardo-netflix (commented):
Looks good overall!

archaius2-core/src/main/java/com/netflix/archaius/config/DefaultSettableConfig.java
kilink: Good catch, I had missed that. I've updated the PR and added tests to cover that.

PR #651, opened 2023-05-19
Property usage instrumentation initial framework

akang31:
Initial framework for capturing property usage in Archaius via an AccessMonitorUtil. 

In this first iteration, AccessMonitorUtil keeps track of a map from PropertyId (new concept) to a wrapper around a List of property usage events (which only contain a timestamp). We are not overly concerned about the granularity of the property usage data at this stage, so we also only keep one such event per iteration (effectively just giving us a map from property id to last used timestamp). We leave the data model here generic enough to augment it with additional metadata at the high level, and property level (e.g. we could also add additional scope-related metadata, maintain a count of property accesses in the future, or keep track of more events per period).

AccessMonitorUtil also has a setDataFlushConsumer method, which when applied will on a (default) 2 minute cadence flush the stored property usage data to the method and reset the existing data. The idea is for this to be used to send to a remote database/stream for aggregation and analysis.

We initially only support instrumenting in the PollingDynamicConfig and propagate the settings through CompositeConfig. This is only because our primary target for instrumentation is remote properties. We should support instrumentation in other dependent configs (LayeredConfig, PrefixedViewConfig, and PrivateViewConfig) and there is nothing stopping us from supporting it in the other configs as well.

There are also changes in JsonPersistedV2Reader to hold onto property ids, and not just keys and values, so that we can uniquely tie usage data with an actual property when the data is passed along for analysis. 

To enable AccessMonitorUtil monitoring in Guice, an AccessMonitorUtil must be bound along with the Persisted2ConfigProvider (this will also enable the JsonPersistedV2Reader reading property ids). To enable data flushing, the bound AccessMonitorUtil must have a DataFlushConsumer set.

In Spring, the idea will be the same, with the presence of an AccessMonitorUtil enabling monitoring, and a DataFlushConsumer being set enabling the flushing.

Followups:
 * Ability to quickly disable flushing
 * Ability to quickly disable instrumentation
 * Conditional flushing (e.g. blocking flushes and warning instead when the property count is too high)


archaius2-api/src/main/java/com/netflix/archaius/api/instrumentation/AccessMonitorUtil.java
rgallardo-netflix, line 1: This probably belongs in the core jar, not in api. The distinction in my mind is that api should have things used by a library that reads config values but doesn't really care where they come from, while core should have the things needed to actually set up a `Config`  object instance for the libraries to get.

rgallardo-netflix, line 1: Just to be a good citizen, make this class an `AutoCloseable` and give it a `close()` method that tries to flush one last time and shuts down the scheduled task and the thread pool.

archaius2-api/src/main/java/com/netflix/archaius/api/instrumentation/AccessMonitorUtil.java
```diff
@@ -0,0 +1,157 @@
...
+        this.dataFlushConsumer = dataFlushConsumer;
+        this.recordStackTrace = recordStackTrace;
+        this.executor = Executors.newSingleThreadScheduledExecutor(
+                runnable -> {
+                    Thread thread = Executors.defaultThreadFactory().newThread(runnable);
+                    thread.setDaemon(true);
+                    thread.setName(String.format("Archaius-Instrumentation-Flusher-%d", counter.incrementAndGet()));
+                    return thread;
+                });
+        startFlushing(initialFlushDelaySeconds, flushPeriodSeconds);
```
rgallardo-netflix: Race condition alert!
The java memory model doesn't really like it when you start a thread inside a constructor. There's a tiny chance that the new thread could see the object in a not-fully-initialized state. To avoid this, it'd be better to move the call to `startFlushing()` to the `Builder.build()` method.

archaius2-api/src/main/java/com/netflix/archaius/api/instrumentation/AccessMonitorUtil.java
```diff
@@ -0,0 +1,157 @@
...
+    public Map<String, PropertyUsageData> getUsageMapImmutable() {
+        return Collections.unmodifiableMap(new HashMap<>(propertyUsageMap));
+    }
+
+    public Map<String, Integer> getStackTracesImmutable() {
+        return Collections.unmodifiableMap(new HashMap<>(stackTrace));
+    }
+
+    public boolean flushingEnabled() {
+        return dataFlushConsumer != null;
```
rgallardo-netflix: I'm assuming this is where we'd disable flushing at runtime?

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -239,7 +239,7 @@ public Config getPrivateView() {
 
     @Override
     public <T> T accept(Visitor<T> visitor) {
-        forEachProperty(visitor::visitKey);
+        forEachPropertyUninstrumented(visitor::visitKey);
```
rgallardo-netflix, line 244: Ugh. Let's see if we can find the users of this, to make sure they really all are platform-level stuff that should not be logged. I can easily see how something like `WebClient` or NIWS could use this to walk through the tree looking for the configs for a specific client instance :-/

archaius2-persisted2/src/test/java/com/netflix/archaius/persisted2/JsonPersistedV2ReaderTest.java
rgallardo-netflix, line 1: Nice! Thanks!

archaius2-core/src/main/java/com/netflix/archaius/config/PollingDynamicConfig.java
```diff
@@ -127,6 +142,42 @@ public Iterable<String> keys() {
...
+            consumer.accept(k, v);
+        });
+    }
+
+    @Override
+    public void forEachPropertyUninstrumented(BiConsumer<String, Object> consumer) {
         current.forEach(consumer);
     }
+
+    public void setAccessMonitorUtil(AccessMonitorUtil accessMonitorUtil) {
```
rgallardo-netflix: Race condition alert!
I assume you kept these two fields separate in order to allow runtime disabling/enabling of instrumentation, right? (if not, just change `instrumentationEnabled()` to just return `accessMonitorUtil != null` and skip the rest of this comment :-) )

These two MUST be set together, or you could end up getting a null pointer in line 175. In order to ensure that without synchronizing everywhere, bundle both fields into a small POJO that's kept in a volatile field. Something like (in record syntax because I'm lazy but we'll need to do it with a normal class b/c java8 :-( )
```
   private static record Instrumentation(AccessMonitorUtil monitor, boolean enabled);
   private volatile Instrumentation instrumentation;

   [...]

  public void setAccessMonitorUtil(monitor) {
     this.instrumentation = new Instrumentation(monitor, true);
  }
```

rgallardo-netflix (changes requested):

archaius2-api/src/main/java/com/netflix/archaius/api/instrumentation/AccessMonitorUtil.java
akang31: Makes sense; updated to be in the core jar instead. PropertyDetails is the only class that needs to remain in api because Config depends on it.

akang31: Updated to be AutoCloseable and have a close() method

archaius2-api/src/main/java/com/netflix/archaius/api/instrumentation/AccessMonitorUtil.java
akang31: Ah I wasn't aware of that, thanks! Updated to setup the AccessMonitorUtil object and then startFlushing in the Builder.build().

archaius2-api/src/main/java/com/netflix/archaius/api/instrumentation/AccessMonitorUtil.java
akang31: Yes, I was thinking to surface a method on AccessMonitorUtil for disabling it.

The exact details of how that would work to avoid runtime impact / infinite loops isn't immediately clear to me (and will probably vary between Spring and Archaius Guice), so I'll address that in a followup.

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
akang31: Visitors appear to be used as they are intended from the instances I've seen (i.e. mostly for logging). There is some usage of the Archaius provided visitors (none of which seem to require instrumentation), and very few additional custom visitors that also don't seem to require instrumentation (e.g. Titus has a CompositeVisitor for doing logging, algo-commons has an endpoint which uses a Visitor to copy every single Property for debugging).

I didn't see any usage for WebClient or NIWS

archaius2-core/src/main/java/com/netflix/archaius/config/PollingDynamicConfig.java
akang31: Ah right, good catch! I updated to use an Instrumentation class to encapsulate both and added a comment for context (also moved the AccessMonitorUtil into the constructor).

akang31 (commented):
Thanks for the review!

akang31 (commented):
Thanks for the review!

akang31 (commented):
Thanks for the review!

akang31 (commented):
Thanks for the review!

akang31 (commented):
Thanks for the review!

archaius2-core/src/main/java/com/netflix/archaius/config/PollingDynamicConfig.java
```diff
@@ -156,11 +163,6 @@ public void forEachPropertyUninstrumented(BiConsumer<String, Object> consumer) {
...
     }
 
-    public void setAccessMonitorUtil(AccessMonitorUtil accessMonitorUtil) {
-        this.accessMonitorUtil = accessMonitorUtil;
-        instrumentationEnabled = true;
-    }
-
     @Override
     public void recordUsage(PropertyDetails propertyDetails) {
         if (instrumentationEnabled()) {
```
rgallardo-netflix, line 170: Let's just put a note here for when we come up with the runtime-disabling mechanism. If it includes a way to change the AccessMonitorUtil object, then we'll still have a window for a race (if it is not null at the point of this check and it gets unset before we get to the `registerUsage` call. In that case we should probably move the body of this method to live inside the `Instrumentation` object, which would get around the possible issue.
But for now it's probably best to leave it like this, it's probably easier to read.

rgallardo-netflix:
Looks good. Have you tried making a local build and running it in an app?

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
rgallardo-netflix: Good! Let's have this explicitly documented as a limitation.

archaius2-core/src/main/java/com/netflix/archaius/config/PollingDynamicConfig.java
akang31: Thanks for the callout, I added a comment here to remind us to ensure that there isn't a race condition when we add this mechanism

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
akang31: Added documentation here and also in the Config.Visitor interface

akang31 (commented):
I've done some local builds to ensure that this is working in both SBN and Archaius2 via Guice.

I found that PrefixedViewConfig is actually used very often, so I went ahead and updated PrefixedViewConfig, PrivateViewConfig, and LayeredConfig to all use a AbstractDependentConfig superclass which relies on data in an immutable CachedState object to handle normal property operations and instrumentation propagation.

PR #663, opened 2023-07-17
Basic instrumentation changes behind a flag

akang31:
Keeps track of usage data through ConcurrentCompositeConfiguration when enabled by the system property and also surfaces an uninstrumented endpoint for any calls that should not be recorded.

Also includes the ability to record stack traces of call sites (useful for debugging, but should not be used in prod).


archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
```diff
@@ -102,8 +104,21 @@
...
     private Map<String, AbstractConfiguration> namedConfigurations = new ConcurrentHashMap<String, AbstractConfiguration>();
-    
+
+    private Map<String, Integer> stackTraces = new ConcurrentHashMap<>();
+
+    public Set<String> getUsedProperties() {
+        return usedProperties;
+    }
+
+    private Set<String> usedProperties = new HashSet<>();
```
rgallardo-netflix: Either this field should be a `ConcurrentHashMap` like the `stackTraces` one, or neither should because they'll both be subjected to the very same possible races.

Also, could they both be `final`?

akang31: Updated usedProperties to also be backed by a ConcurrentHashMap and changed both fields to be final as well.

PR #670, opened 2023-08-31
Fix debug logs in ConfigProxyFactory

kilink:
Fix debug logs in createInterfaceProperty / createScalarProperty to actually log the relevant types, instead of always logging "Class" and "Proxy".

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
@@ -16,7 +16,6 @@ import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
 import java.lang.invoke.CallSite;
-import java.lang.invoke.LambdaConversionException;
 import java.lang.invoke.LambdaMetafactory;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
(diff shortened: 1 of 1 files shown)


archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -377,13 +376,13 @@ private static <T> Function<Object[], T> createDefaultMethodSupplier(Method meth
         };
     }
 
-    protected <T> MethodInvoker<T> createInterfaceProperty(String propName, final T proxy) {
-        LOG.debug("Creating interface property `{}` for type `{}`", propName, proxy.getClass());
+    protected <T> MethodInvoker<T> createInterfaceProperty(String propName, final T proxy, Class<?> targetClass) {
```
rgallardo-netflix, line 379: So, basically, this method exists just so that 1) we have a neat place to log, and 2) the internal API remains symmetrical with scalar and parametrized properties?
I'm not objecting, just making sure I'm getting it :-)

kilink: oh... actually now that I see it's a protected method, I likely shouldn't change the signature at all. I can move the logging outside of the method I suppose.

PR #672, opened 2023-08-31
Enable running tests in JDKs != 8

rgallardo-netflix:
This required adding removing dependencies in packages that were removed in JDK9 and later.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,7 @@ subprojects {
     }
 
     tasks.withType(Test) {
-        var testJdk = System.getProperty("testJdk", "17")
+        var testJdk = System.getProperty("testJdk", "8")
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -37,4 +37,12 @@ subprojects {
     tasks.withType(Javadoc) {
       options.addStringOption('Xdoclint:none', '-quiet')
     }
+
```
rgallardo-netflix, line 40: @kilink , try pulling this change into your branch and run with 
```
./gradlew  -DtestJdk=17 clean test
```
to run the tests under 17.
As of this change, I still have a bunch of failures, but they seem to mostly be related to dep mismatches with Mockito :-(

PR #680, opened 2023-09-28
Update handling for parameterized methods via ConfigProxyFactory

### LINKED ISSUE #676: @Configuration(prefix = "foo") on an interface, in conjunction with a parameterized @PropertyName fails
When applying @Configuration(prefix = "foo") on an interface, we would expect that any @PropertyNames specified in the interface would automatically look for foo.[names], but this does not apply when there are parameters specified, instead using the PropertyName directly.

Example

```
@Configuration(prefix = "foo") 
interface BogusConfiguration {
  @DefaultValue("BAD")
  @PropertyName(name = "bar")
  String getFooBar();

  @DefaultValue("BAD")
  @PropertyName(name = "${0}")
  String getFoo(String bar);
}
```

Given config
```
foo.bar: "GOOD"
foo.baz: "ALSO GOOD"
bar: "ALSO BAD"
```

Expected
```
getFooBar() --> GOOD
getFoo("bar") --> GOOD
getFoo("baz") --> ALSO GOOD
```

Actual
```
getFooBar() --> GOOD
getFoo("bar") --> ALSO BAD
getFoo("baz") --> BAD // From the default value
```

Issue is in ConfigProxyFactory.createParameterizedProperty.

Changing this to only the correct behavior would be breaking for many clients, we will include backward compatibility for cases in which clients specify the prefix (e.g. in the above example, a name = "foo.${0}" would also produce the same result). 

This would however still be breaking for clients that specified a prefix at the top level config, and then ignored it in the body.

akang31:
Currently, even if a top level prefix is specified in an @Configuration annotation, parameterized methods in the body will ignore the prefix.

This change updates this behavior to properly acknowledge the prefix annotation.

This is notably breaking, so we keep backward compatibility for users who compensated by having their parameterized methods include the prefix. However, we will be breaking users that did not include the prefix.

Putting this up for review now, but still following up with a few clients to ensure the safety of this change.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
@@ -9,6 +9,7 @@ import com.netflix.archaius.api.annotations.Configuration;
 import com.netflix.archaius.api.annotations.DefaultValue;
 import com.netflix.archaius.api.annotations.PropertyName;
 import com.netflix.archaius.util.Maps;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.SystemUtils;
 import org.apache.commons.lang3.text.StrSubstitutor;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


rgallardo-netflix:
Looks good, let's wait for those users' feedback.

akang31:
I've received confirmation from the users that this change will not affect them (the properties were either unused, or in decommissioned applications)

rgallardo-netflix:
👍 Let's merge, then. But also, since there's a visible behavior change, let's remember to bump the next release's minor version to 2.6

PR #686, opened 2023-10-11
Lambdas for proxies take2

rgallardo-netflix:
This PR re-introduces the usage of lambdas for config proxies. I refactored things into a separate class and split the implementations for different JVMs to make things cleaner. I also introduced a "usage counter" circuit breaker to reduce the chance that users could hit OOM errors like they did before. I'll probably end up removing that later, thought.

There's also a bunch of compiler warning cleanups. I tried to keep those on a separate commit so the actual change is easier to review.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/DefaultMethodInvokerFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/DefaultMethodInvokerFactory.java
@@ -1,0 +1,238 @@
+package com.netflix.archaius;
+
+import org.apache.commons.lang3.JavaVersion;
+import org.apache.commons.lang3.SystemUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.invoke.CallSite;
+import java.lang.invoke.LambdaMetafactory;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+abstract class DefaultMethodInvokerFactory {
+    private static final Logger LOG = LoggerFactory.getLogger(DefaultMethodInvokerFactory.class);
+
+    private static final DefaultMethodInvokerFactory INSTANCE;
+    static {
+        if (SystemUtils.isJavaVersionAtMost(JavaVersion.JAVA_1_8)) {
+            INSTANCE = new LegacyJava8Factory();
+        } else if (SystemUtils.isJavaVersionAtMost(JavaVersion.JAVA_16)) {
+            INSTANCE = new BoundMethodHandlesFactory();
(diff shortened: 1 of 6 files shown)


rgallardo-netflix:
I've also realized that, since a ConfigProxyFactory is always initialized *after* the config object it proxies ... we could have a flag to disable use of lambdas! So anyone hit by the weirdness can just set the flag and keep on leaking proxies to their hearts contents 😁

archaius2-core/src/main/java/com/netflix/archaius/DefaultMethodInvokerFactory.java
```diff
@@ -0,0 +1,238 @@
...
+    private static final DefaultMethodInvokerFactory INSTANCE;
+    static {
+        if (SystemUtils.isJavaVersionAtMost(JavaVersion.JAVA_1_8)) {
+            INSTANCE = new LegacyJava8Factory();
+        } else if (SystemUtils.isJavaVersionAtMost(JavaVersion.JAVA_16)) {
+            INSTANCE = new BoundMethodHandlesFactory();
+        } else { // Java 17 onwards
+            INSTANCE = new LambdasFactory();
+        }
+        LOG.info("Choosing {} as invoker factory for default methods", INSTANCE.getClass());
```
codefromthecrypt, line 33: is info or debug more sensible for this?

archaius2-core/src/main/java/com/netflix/archaius/DefaultMethodInvokerFactory.java
```diff
@@ -0,0 +1,238 @@
...
+        // We keep a running count of how many times we've seen a given Class. If we cross the threshold we stop creating
+        // lambdas for that class and fall back to the reflective implementation. This is motivated by an edge case we
+        // saw in clients, where they create (and leak) large numbers of proxies to the same interface. Because each lambda
+        // object requires its own anon class object, which lives in metaspace, and since users sometimes size their
+        // metaspaces to be much smaller than the heap, the leak becomes much more visible and causes OOMs.
+        private static final Map<String, Integer> SEEN_COUNTS = new ConcurrentHashMap<>();
+
+        // The threshold is simply a best guess. Creating 1000 lambdas for the same interface is almost certainly
+        // a leak, right?
+        private static final int MAX_SEEN_THRESHOLD = 1000;
```
codefromthecrypt, line 98: would a JMH bench trigger stuff like this?

archaius2-core/src/main/java/com/netflix/archaius/DefaultMethodInvokerFactory.java
```diff
@@ -0,0 +1,238 @@
...
+
+     }
+
+
+    /**
+     * For Java 17 onwards, we wrap the method calls in lambdas and return that. This skips the reflection machinery and
+     * provides a noticeable performance boost. This implementation is known to fail in java 9 - 11. It *should* work
+     * from 12 or 13 onwards, but we have only tested it in 17 and 21.
+     */
+    private static class LambdasFactory extends DefaultMethodInvokerFactory {
```
codefromthecrypt, line 88: if package-scoped (final, too) you could make a matrix test and benchmarks, or even fuzzing to rely less on seen in usage stuff.

codefromthecrypt (commented):
I remember loving this project, and randomly took a look since such an interesting situation. Pay no mind to my unsolicited 2p unless you care to!

PR #687, opened 2023-10-20
Add utility methods to simplify usage of collection-valued properties.

### LINKED ISSUE #654: Archaius should provide some simple syntax to describe the parametrized collection types that are decodable by default.
`AbstractConfig#getValueWithDefault(Class, String, T)` is NOT able to decode collection types, but `AbstractConfig#(Type, String, T)` can, if a `ParametrizedType` is passed. We should provide a shorthand method to create such `ParametrizedTypes`. Otherwise, users are forced to hunt themselves for a way to create them.

Apache Commons Lang 3's `TypeUtils` has sample code that we could copy and specialize for this restricted use case.

rgallardo-netflix:
Although Archaius has supported decoding of collection-valued properties such as list, sets and maps since at least 2019, the syntax to use them was never properly documented and required users to provide their own implementation of the ParametrizedType interface.

This commit adds some syntactic sugar to simplify usage.

Closes: #654

### DIFF
--- a/archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
@@ -1,0 +1,95 @@
+package com.netflix.archaius.api;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


akang31 (approved):
This clarifies this very well. Thanks for the change!

PR #688, opened 2023-10-20
Readability refactor

rgallardo-netflix:
Refactored code for readability. Split out most sections into self-contained methods with a single objective. Code is now longer, but less convoluted and with less anonymous nested things.

Cleaned up compiler warnings in main class and in the corresponding unit tests.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
@@ -104,7 +104,17 @@ public class ConfigProxyFactory {
     private final Decoder decoder;
     private final PropertyRepository propertyRepository;
     private final Config config;
-    
+
+
+    /**
+     * Build a proxy factory from the provided config, decoder and PropertyFactory. Normal usage from most applications
+     * is to just set up injection bindings for those 3 objects and let your DI framework find this constructor.
+     *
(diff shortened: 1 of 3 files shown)


archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -182,131 +180,145 @@ interface MethodInvoker<T> {
...
-                    sbProperty.append("'");
-                    return sbProperty.toString();
-                }).collect(Collectors.joining(",")));
-                sb.append("]");
-                return sb.toString();
+            propertyNames.put(method, methodInvokerHolder.propertyName);
+
+            if (immutable) {
+                // Cache the current value of the property and always return that.
+                // Note that this will fail for parametrized properties!
```
akang31: There are a few instances where parameterized is misspelled parametrized

akang31 (approved):
Thanks for the changes! This makes ConfigProxyFactory much more self-explanatory.

archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
akang31: Oh it looks like both are valid, but I think it'd be best if we stuck with one or the other regardless

rgallardo-netflix: LOL, thanks for catching this!

PR #690, opened 2023-10-30
Log "excessive" creation of factories and proxies

rgallardo-netflix:
Add logging to help users find leaks of Config proxies and their factories.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
@@ -16,6 +16,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PrintWriter;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
@@ -24,6 +27,7 @@ import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
@@ -31,6 +35,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.WeakHashMap;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -98,6 +103,12 @@ import java.util.stream.Collectors;
 public class ConfigProxyFactory {
     private static final Logger LOG = LoggerFactory.getLogger(ConfigProxyFactory.class);
 
+    // Users sometimes leak both factories and proxies, leading to memory problems that get blamed on us ;-)
+    // We'll use these maps to keep track of how many instances of each are created and make log noise to help them
+    // track down the actual culprits. WeakHashMaps to avoid holding onto objects ourselves.
+    private static final Map<Config, Integer> FACTORIES_COUNT = Collections.synchronizedMap(new WeakHashMap<>());
+    private static final Map<Class<?>, Integer> PROXIES_COUNT = Collections.synchronizedMap(new WeakHashMap<>());
+
     /**
      * The decoder is used for the purpose of decoding any @DefaultValue annotation
      */
@@ -121,6 +132,8 @@ public class ConfigProxyFactory {
         this.decoder = decoder;
         this.config = config;
         this.propertyRepository = factory;
+
+        warnWhenTooMany(FACTORIES_COUNT, config, 5, String.format("ProxyFactory(Config:%s)", config.hashCode()));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -460,6 +475,29 @@ private static void maybeWrapThenRethrow(Throwable t) {
...
+    private static <T> void warnWhenTooMany(Map<T, Integer> counters, T countKey, int limit, String objectDescription) {
+        int currentCount = counters.merge(countKey, 1, Integer::sum);
+
+        // Emit warning if we're over the limit BUT only when the current count is a multiple of the limit :-)
+        // This is to avoid being *too* noisy
+        if (LOG.isWarnEnabled() &&
+            currentCount >= limit &&
+            (currentCount % limit == 0 )) {
+
+                String stackTraceForDebuggint;
```
iuliiasobolevska: nit: typo in `stackTraceForDebuggint`? (debugging)

archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -189,6 +202,8 @@ interface PropertyValueGetter<T> {
 
     @SuppressWarnings({"unchecked", "rawtypes"})
     <T> T newProxy(final Class<T> type, final String initialPrefix, boolean immutable) {
+        warnWhenTooMany(PROXIES_COUNT, type, 50, String.format("Proxy(%s)", type));
```
iuliiasobolevska: would it make sense to make magic constants configurable?

archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -98,6 +103,12 @@
 public class ConfigProxyFactory {
     private static final Logger LOG = LoggerFactory.getLogger(ConfigProxyFactory.class);
 
+    // Users sometimes leak both factories and proxies, leading to memory problems that get blamed on us ;-)
```
iuliiasobolevska: IMHO it would be nice to rewrite the message in lieu of the "true" Netflix culture instead of continuing "blame" culture :)

archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -98,6 +103,12 @@
...
+    // We'll use these maps to keep track of how many instances of each are created and make log noise to help them
```
iuliiasobolevska: extra nit: code style: different javadoc style from the rest of the file

rgallardo-netflix: Not really a javadoc, more an internal comment for implementors :-)

archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
akang31: 50 seems a little high for this in general, is this to accommodate the possibility of someone having the same proxy class but with different prefixes? If so, would it make sense to combine that into the key?

rgallardo-netflix: Trick here is that some of this code could be running before config is fully initialized. Let me think it through.

rgallardo-netflix: Yeah, that's the thought.

archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -460,6 +485,19 @@ private static void maybeWrapThenRethrow(Throwable t) {
...
+    private static <T> void warnWhenTooMany(Map<T, Integer> counters, T countKey, int limit, Supplier<String> objectDescription) {
+        int currentCount = counters.merge(countKey, 1, Integer::sum);
+
+        // Emit warning if we're over the limit BUT only when the current count is a multiple of the limit :-)
+        // This is to avoid being *too* noisy
+        if (LOG.isWarnEnabled() &&
+            currentCount >= limit &&
+            (currentCount % limit == 0 )) {
+
+                LOG.warn("Too many {} objects are being created. Please review the calling code to prevent memory leaks. Stack trace for debugging follows:",
```
iparadiso: Nit: Perhaps log the `EXCESSIVE_PROXIES_LIMIT` knob so users know how to adjust this if they have a legit reason to create so many of these proxies.  But also log that you should only be creating a singleton.

iparadiso (approved):

akang31 (approved):

archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
rgallardo-netflix: Done :-)
New log message looks like
```
2023-10-30 13:30:48,612 WARN  ConfigProxyFactory:497  Too many ProxyFactory(Config:330084561) objects are being created (5 so far).
Please review the calling code to prevent memory leaks.
Normal usage for ConfigProxyFactory is to create singletons via your DI mechanism.
For special use cases that *require* creating multiple instances you can tune reporting
by setting the `archaius.excessiveProxiesLogging.limit` config key to a higher threshold.
Stack trace for debugging follows: [Test worker]
java.lang.Throwable
	at com.netflix.archaius.ConfigProxyFactory.warnWhenTooMany(ConfigProxyFactory.java:503)
	at com.netflix.archaius.ConfigProxyFactory.<init>(ConfigProxyFactory.java:146)
	at com.netflix.archaius.ProxyFactoryTest.testLogExcessiveUse(ProxyFactoryTest.java:563)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        ...
```

PR #691, opened 2023-11-19
Return unmodifiable collections in DefaultCollectionsTypeConverterFactory

kilink:
DefaultCollectionsTypeConverterFactory already returned an unmodifiable Map, so update the collection converters for List, Set, and SortedSet to also wrap the converted result with the appropriate unmodifiable collection. Additionally, if the value to convert is an empty String, don't allocate a collection and simply return the appropriate empty singleton (e.g. Collections.emptyList()).

Finally, improve type safety in a few places and avoid usage of raw types.

### DIFF
--- a/archaius2-core/src/test/java/com/netflix/archaius/property/PropertyTest.java
+++ b/archaius2-core/src/test/java/com/netflix/archaius/property/PropertyTest.java
@@ -19,6 +19,7 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


archaius2-core/src/test/java/com/netflix/archaius/property/PropertyTest.java
```diff
@@ -167,6 +168,43 @@ public void testCollectionTypes() {
         // Test proper handling of unset properties
         Property<Map<CustomType, CustomType>> emptyProperty = factory.getMap("fubar", CustomType.class, CustomType.class);
         Assert.assertNull(emptyProperty.get());
+
+        config.setProperty("fubar", "");
+        Property<List<String>> emptyListProperty = factory.getList("fubar", String.class);
+        Assert.assertEquals(Collections.emptyList(), emptyListProperty.get());
```
rgallardo-netflix, line 174: Since all the default empty collections are value types, this assertion is really just equivalent to `assertTrue(emptyList.isEmpty())`, isn't it?

kilink: Yeah, that's equivalent.

PR #695, opened 2023-12-08
Allow conversion for interface types

kilink:
Change the behavior of ConfigProxyFactory to check the Decoder for the existence of a TypeConverter before assuming that an interface type should be a nested archaius proxy.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
@@ -300,8 +300,8 @@ public class ConfigProxyFactory {
             if (!knownCollections.containsKey(returnType)
                     && returnType.isInterface()
                     && !(decoder instanceof TypeConverter.Registry && ((TypeConverter.Registry) decoder).get(m.getGenericReturnType()).isPresent())) {
-                // Our return type is an interface but not a known collection. We treat it as a nested Config proxy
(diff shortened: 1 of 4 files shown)


archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
```diff
@@ -296,7 +297,9 @@ private <T> MethodInvokerHolder buildInvokerForMethod(Class<T> type, String pref
             // This object encapsulates the way to get the value for the current property.
             final PropertyValueGetter propertyValueGetter;
 
-            if (!knownCollections.containsKey(returnType) && returnType.isInterface()) {
+            if (!knownCollections.containsKey(returnType)
+                    && returnType.isInterface()
+                    && !(decoder instanceof TypeConverter.Registry && ((TypeConverter.Registry) decoder).get(m.getGenericReturnType()).isPresent())) {
                 // Our return type is an interface but not a known collection. We treat it as a nested Config proxy
```
rgallardo-netflix: Update comment here, please :-)

rgallardo-netflix (approved):

archaius2-core/src/main/java/com/netflix/archaius/ConfigProxyFactory.java
kilink: I updated the comment.

PR #698, opened 2024-01-17
Update instrumentation with more call sites and instrumentation clearing endpoint

akang31:
The instrumentation currently is missing a few cases that are addressed here:

- OverrideProperties were not instrumented
- Calls through getStringArray (which for NetflixConfiguration came via the getString call on for both DynamicProperty wrappers [updateValue] and direct NetflixConfiguration calls, covering a large portion of calls [basically anything not calling getProperty])

We also add the ability to clear used instrumentation to avoid repeatedly sending the same instrumentation data over the course of an application's lifetime.

### DIFF
--- a/archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
+++ b/archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
@@ -546,6 +546,9 @@ public class ConcurrentCompositeConfiguration extends ConcurrentMapConfiguration
     private Object getProperty(String key, boolean instrument)
     {
         if (overrideProperties.containsKey(key)) {
+            if (instrument) {
+                recordUsage(key);
+            }
             return overrideProperties.getProperty(key);
         }
         Configuration firstMatchingConfiguration = null;
@@ -922,12 +925,13 @@ public class ConcurrentCompositeConfiguration extends ConcurrentMapConfiguration
      * @param config the configuration to query
      * @param key the key of the property
      */
-    private static void appendListProperty(List<Object> dest, Configuration config,
+    private void appendListProperty(List<Object> dest, Configuration config,
             String key)
     {
         Object value = config.getProperty(key);
         if (value != null)
         {
+            recordUsage(key);
             if (value instanceof Collection)
             {
                 Collection<?> col = (Collection<?>) value;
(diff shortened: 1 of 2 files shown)


archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
```diff
@@ -115,9 +115,16 @@ public class ConcurrentCompositeConfiguration extends ConcurrentMapConfiguration
     private final Set<String> usedProperties = ConcurrentHashMap.newKeySet();
 
     public Set<String> getUsedProperties() {
-        return usedProperties;
+        return Collections.unmodifiableSet(new HashSet<>(usedProperties));
     }
 
+    public Set<String> getAndClearUsedProperties() {
+        synchronized (usedProperties) {
```
rgallardo-netflix: Synchronizing on the map won't keep it from being modified concurrently by other threads, so there still a race condition where threadA calls getUsedProperties, threadB records a new usage, threadA clears the map, and now threadB's usage is lost.
The only way to avoid that would be to synchronize all methods that record access, in addition to this one. Not sure that'd be worth the performance hit, but in that case you might as well change the map back to a regular HashMap, instead of paying the ConcurrentHashMap tax.

akang31: Per offline discussion, updated this to use an AtomicReference instead

archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
```diff
@@ -109,15 +110,19 @@ public class ConcurrentCompositeConfiguration extends ConcurrentMapConfiguration
...
+    private final AtomicReference<Set<String>> usedPropertiesRef = new AtomicReference<>(ConcurrentHashMap.newKeySet());
 
     public Set<String> getUsedProperties() {
-        return usedProperties;
+        return Collections.unmodifiableSet(new HashSet<>(usedPropertiesRef.get()));
     }
 
+    public Set<String> getAndClearUsedProperties() {
+        Set<String> ret = usedPropertiesRef.getAndSet(ConcurrentHashMap.newKeySet());
+        return Collections.unmodifiableSet(new HashSet<>(ret));
```
rgallardo-netflix: Same comment about double wrapping as in the other PR.

akang31: Removed the double wrap here to just directly return the unmodifiableSet

rgallardo-netflix (approved):

PR #699, opened 2024-01-19
Fix race condition on propertyUsageMap via AtomicReference

akang31:
Address the race condition on propertyUsageMap when we are trying to flush the existing map and then empty it. We use the AtomicReference here to hot swap the existing map out for a new empty one before we then copy the old map to be flushed.

Also tacked on a small logging change here, which flips the logging to only happen when it starts up the flushing and doesn't print at all when nothing is flushing. SBN by default will setup the AccessMonitorUtil twice, so we just add confusing noise for no reason with the prior logs.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
@@ -139,8 +139,7 @@ public class AccessMonitorUtil implements AutoCloseable {
     }
 
     private Map<String, PropertyUsageData> getAndClearUsageMap() {
-        Map<String, PropertyUsageData> map =
-                propertyUsageMapRef.getAndUpdate(unused -> new ConcurrentHashMap<>());
+        Map<String, PropertyUsageData> map = propertyUsageMapRef.getAndSet(new ConcurrentHashMap<>());
         return Collections.unmodifiableMap(new HashMap<>(map));
     }
 
--- a/archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
@@ -16,6 +16,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
 /** Tracks property usage data and flushes the data periodically to a sink. */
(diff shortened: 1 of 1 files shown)


archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
```diff
@@ -138,15 +139,12 @@ private List<PropertyUsageEvent> createEventList(PropertyUsageEvent event) {
     }
 
     private Map<String, PropertyUsageData> getAndClearUsageMap() {
-        synchronized (propertyUsageMap) {
-            Map<String, PropertyUsageData> ret = getUsageMapImmutable();
-            propertyUsageMap.clear();
-            return ret;
-        }
+        Map<String, PropertyUsageData> map = propertyUsageMapRef.getAndSet(new ConcurrentHashMap<>());
+        return Collections.unmodifiableMap(new HashMap<>(map));
```
rgallardo-netflix: No need to double wrap here. If any other thread had a reference to the map they'll release fairly soon anyway. Making a copy here will lose the updates those threads were recording. If you simply wrap it in the `unmodifiableMap` then there's a little bit more time between here and the actual flush to give them a chance to finish.

archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
```diff
@@ -108,8 +109,8 @@ private void flushUsageData() {
 
     /** Merge the results of given accessMonitorUtil into this one. */
     public void merge(AccessMonitorUtil accessMonitorUtil) {
-        for (Map.Entry<String, PropertyUsageData> entry : accessMonitorUtil.propertyUsageMap.entrySet()) {
-            propertyUsageMap.putIfAbsent(entry.getKey(), entry.getValue());
+        for (Map.Entry<String, PropertyUsageData> entry : accessMonitorUtil.propertyUsageMapRef.get().entrySet()) {
+            propertyUsageMapRef.get().putIfAbsent(entry.getKey(), entry.getValue());
```
rgallardo-netflix: Grab a reference to the destination map before entering the loop and merge the entries into that. Otherwise, if a call to `getAndClear` comes in while the loop is running then half of the other `accessMonitorUtil`'s entries will be merged into the old map and half into the new, which will be confusing at best :-P

archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
akang31: Ah that makes sense, updated to this

archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
akang31: Oops yea good point. I think in practice this should never happen, but good to safeguard against it

rgallardo-netflix (approved):

PR #700, opened 2024-02-04
Add equality comparison for ArchaiusType

gavinbunney:
This PR adds equality methods for the `ArchaiusType` class to ensure `keyAndType` references are reused correctly within the `DefaultPropertyFactory`.

The [`computeIfAbsent`](https://github.com/Netflix/archaius/blob/2.x/archaius2-core/src/main/java/com/netflix/archaius/DefaultPropertyFactory.java#L194) check for existing property references uses the [`type` equality](https://github.com/Netflix/archaius/blob/2.x/archaius2-core/src/main/java/com/netflix/archaius/DefaultPropertyFactory.java#L358-L362) within `keyAndType` to see if a reference to a property already exists, otherwise it adds it to the property ref map.

This can cause an explosion of allocations where the same property key and type are supplied, but as the type itself is constructed on every fetch (`new ArchaiusType`), the default java object equals methods always return false, so a new reference is added to the property factory.

### DIFF
--- a/archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
@@ -92,4 +92,38 @@ public class ArchaiusType implements ParameterizedType {
         String typeArgumentNames = Arrays.stream(typeArguments).map(Class::getSimpleName).collect(Collectors.joining(","));
         return String.format("parameterizedType for %s<%s>", rawType.getSimpleName(), typeArgumentNames);
     }
+
+    @Override
+    public int hashCode() {
+        int result = 1;
+        result = 31 * result + (this.rawType == null ? 0 : this.rawType.hashCode());
+        result = 31 * result + Arrays.hashCode(this.typeArguments);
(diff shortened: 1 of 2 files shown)


sullis:
Suggested enhancement:  use the [equalsverifier library](https://github.com/jqno/equalsverifier) to confirm the behavior of the **equals** method.

Add the library to build.gradle:
```
testImplementation 'nl.jqno.equalsverifier:equalsverifier:3.+'
```

And add a unit test:

```
package com.netflix.archaius.api;

import nl.jqno.equalsverifier.EqualsVerifier;
import org.junit.Test;

public class EqualsVerificationTest {
    @Test
    public void verifyArchaiusType() {
        verifyEquals(ArchaiusType.class);
    }

    private static void verifyEquals(Class<?> clazz) {
        EqualsVerifier.forClass(clazz)
                .usingGetClass()
                .verify();
    }

}

```

rgallardo-netflix:
@sullis, I like that idea but I think I'd rather do it separately as a bigger change adding verifications to every class with an `equals` method. I did test things out locally and the PR passes as it is now.

PR #701, opened 2024-02-06
Improve Stack Trace tracking for Instrumentation

akang31:
Add archaius_stack_trace_enabled_properties which allows tracking of individual properties (instead of all of them) so we can reduce runtime impact of calculating stack traces. We can use this to only track some unused global fast properties.

This new set will also maintain all the properties associated with specific call sites, in case we are calling several.

Also surfaces these Maps as they were only accessible via the debugger before

### DIFF
--- a/archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
+++ b/archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
@@ -116,6 +116,7 @@ public class ConcurrentCompositeConfiguration extends ConcurrentMapConfiguration
     private Map<String, AbstractConfiguration> namedConfigurations = new ConcurrentHashMap<>();
 
     private final Map<String, Integer> stackTraces = new ConcurrentHashMap<>();
+    private final Map<String, Set<String>> stackTracesAndProperties = new ConcurrentHashMap<>();
     private final AtomicReference<Set<String>> usedPropertiesRef = new AtomicReference<>(ConcurrentHashMap.newKeySet());
 
     private Set<String> convertStringFlag(String properties) {
@@ -138,6 +139,11 @@ public class ConcurrentCompositeConfiguration extends ConcurrentMapConfiguration
 
     public Map<String, Integer> getInstrumentationStackTraces() {
         return Collections.unmodifiableMap(new HashMap<>(stackTraces));
+    }
+
+    public Map<String, Set<String>> getInstrumentationStackTracesAndProperties() {
+        // Shallow copy
+        return Collections.unmodifiableMap(new HashMap<>(stackTracesAndProperties));
     }
 
     private List<AbstractConfiguration> configList = new CopyOnWriteArrayList<AbstractConfiguration>();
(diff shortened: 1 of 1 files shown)


archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
```diff
@@ -589,13 +611,30 @@ private Object getProperty(String key, boolean instrument)
     public void recordUsage(String key) {
         if (enableInstrumentation) {
             usedPropertiesRef.get().add(key);
-            if (enableStackTrace) {
+            boolean isTrackedProperty = !stackTraceEnabledProperties.isEmpty() && stackTraceEnabledProperties.contains(key);
```
rgallardo-netflix: No need to check separately if the set is empty. A call to `Collections.emptySet().contains()` is just as fast as `Collections.emptySet().isEmpty()`

archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
```diff
@@ -107,14 +108,26 @@ public class ConcurrentCompositeConfiguration extends ConcurrentMapConfiguration
...
+    private final Set<String> stackTraceEnabledProperties = convertStringFlag(System.getProperty(STACK_TRACE_ENABLED_PROPERTIES));
 
     private Map<String, AbstractConfiguration> namedConfigurations = new ConcurrentHashMap<>();
 
     private final Map<String, Integer> stackTraces = new ConcurrentHashMap<>();
+    private final Map<String, Set<String>> stackTracesAndProperties = new ConcurrentHashMap<>();
     private final AtomicReference<Set<String>> usedPropertiesRef = new AtomicReference<>(ConcurrentHashMap.newKeySet());
 
+    private Set<String> convertStringFlag(String properties) {
+        if (properties == null) {
```
rgallardo-netflix, line 125: Since we already have Guava available here, let's leverage its `Splitter` and `ImmutableSet` classes:
```
if (properties == null) {
 return Collections.emptySet();
}

return ImmutableSet.copyOf(Splitter.on(',').trimResults().omitEmptyStrings().split(properties));
```
This removes the possible confusion that would come from people putting spaces or line breaks in the system property and uses a read-only set implementation with a very very efficient `contains()`.

archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
```diff
@@ -589,13 +611,30 @@ private Object getProperty(String key, boolean instrument)
     public void recordUsage(String key) {
         if (enableInstrumentation) {
             usedPropertiesRef.get().add(key);
-            if (enableStackTrace) {
+            boolean isTrackedProperty = !stackTraceEnabledProperties.isEmpty() && stackTraceEnabledProperties.contains(key);
+            if (enableStackTrace || isTrackedProperty) {
                 String trace = Arrays.toString(Thread.currentThread().getStackTrace());
-                stackTraces.merge(trace, 1, (v1, v2) -> v1 + 1);
+                if (enableStackTrace) {
+                    stackTraces.merge(trace, 1, (v1, v2) -> v1 + 1);
```
rgallardo-netflix, line 619: duplicate?

rgallardo-netflix (changes requested):

archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
akang31: How I've missed ImmutableSet.of()... :') 

Thanks for pointing this out, updated to this

archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
akang31: Ah right, I thought this was a HashSet for some reason, removed the empty check

archaius-core/src/main/java/com/netflix/config/ConcurrentCompositeConfiguration.java
akang31: Resolved offline

rgallardo-netflix (approved):

PR #702, opened 2024-02-13
Add a Decoder that accepts custom TypeConverters

kilink:
Add a new Decoder, `CustomDecoder`, which allows users to easily specify custom `TypeConverter` implementations. `CustomDecoder` exposes a single static factory method, `create`, which accepts a Collection of `TypeConverter.Factory` instances.

As part of adding this feature, refactor `DefaultDecoder` in a re-usable base class, `AbstractRegistryDecoder`, which `CustomDecoder` also extends from.

### DIFF
--- a/archaius2-core/src/test/java/com/netflix/archaius/ProxyFactoryTest.java
+++ b/archaius2-core/src/test/java/com/netflix/archaius/ProxyFactoryTest.java
@@ -4,7 +4,6 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.nullValue;
 
-import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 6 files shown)


archaius2-core/src/test/java/com/netflix/archaius/ProxyFactoryTest.java
```diff
@@ -593,31 +593,18 @@ interface CustomObject {
...
-                }
-                return DefaultDecoder.INSTANCE.get(type);
+        TypeConverter.Factory customTypeConverterFactory = (type, registry) -> {
+            if (type.equals(ConfigWithNestedInterface.CustomObject.class)) {
+                return Optional.of(customObjectTypeConverter);
             }
-        }
+            return Optional.empty();
+        };
+        DefaultDecoder customDecoder = DefaultDecoder.newCustomDecoder(Collections.singletonList(customTypeConverterFactory));
```
rgallardo-netflix: Naming here is ... unfortunate. Do you feel like it would be worthwhile to rename the classes a bit? Maybe move the existing code to a `CustomizableDecoder` that has your static factory and making `DefaultDecoder` be a sublcass of it that calls the constructor with an empty factory list?

kilink: I would probably just create a base class that they both share in that case, `AbstractDecoder` / `AbstractRegistryDecoder`, or something like that.

PR #703, opened 2024-02-22
Avoid class cast exceptions in array and enum converters

kilink:
If a ParameterizedType was passed to ArrayTypeConverterFactory or EnumTypeConverterFactory, a ClassCastException would be thrown when it attempted to unconditionally cast it to Class; update the factories to handle this more gracefully by just returning Optional.empty().

Additionally, update ArrayTypeConverterFactory to handle arrays of primitives, and add unit tests for Enums, arrays, and collections.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/converters/ArrayTypeConverterFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/converters/ArrayTypeConverterFactory.java
@@ -5,6 +5,7 @@ import com.netflix.archaius.api.TypeConverter;
 import java.lang.reflect.Array;
 import java.lang.reflect.Type;
 import java.util.Optional;
+import java.util.function.ObjIntConsumer;
 
 public final class ArrayTypeConverterFactory implements TypeConverter.Factory {
     public static final ArrayTypeConverterFactory INSTANCE = new ArrayTypeConverterFactory();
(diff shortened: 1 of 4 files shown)


rgallardo-netflix (approved):

rgallardo-netflix:
As a curiosity, does the existing code not handle primitive arrays at all, or does it handle them inefficiently, with boxing convertions?

kilink:
> As a curiosity, does the existing code not handle primitive arrays at all, or does it handle them inefficiently, with boxing convertions?

It didn't handle them at all, which I found out when I started adding some unit tests to cover some missing conversions. We could also decide to not support primitive arrays if you think it's not needed. I guess it's not really used at all if no one ever complained, or maybe they hit it and switch to the boxed type?

Here's the exception you would hit if you're curious:

```java
DefaultDecoder.INSTANCE.decode(int[].class, "1,2,3,4");
```

```bash
com.netflix.archaius.exceptions.ParseException: Error decoding type `int[]`
	at com.netflix.archaius.AbstractRegistryDecoder.decode(AbstractRegistryDecoder.java:51)
	at com.netflix.archaius.DefaultDecoder.decode(DefaultDecoder.java:29)
	at com.netflix.archaius.AbstractRegistryDecoder.decode(AbstractRegistryDecoder.java:35)
	at com.netflix.archaius.DefaultDecoder.decode(DefaultDecoder.java:29)
       ...
Caused by: java.lang.ClassCastException: [I cannot be cast to [Ljava.lang.Object;
	at com.netflix.archaius.converters.ArrayTypeConverterFactory.lambda$create$1(ArrayTypeConverterFactory.java:33)
	at com.netflix.archaius.AbstractRegistryDecoder.decode(AbstractRegistryDecoder.java:49)
```

rgallardo-netflix:
I'll merge this with the primitives support. The question was more about how to word the release notes :-)

PR #704, opened 2024-02-24
Make ArchaiusType equals / hashCode compatible with ParameterizedTypeImpl

kilink:
Make ArchaiusType's equals / hashCode / toString match what JDK's ParameterizedTypeImpl does, so that equivalent ParameterizedType instances hash and compare equal with ArchaiusType. Update toString to match what ParameterizedTypeImpl does as well.

Additionally, add do a defensive copy of the returned array in getActualTypeArguments, and add basic unit tests.

### DIFF
--- a/archaius2-api/build.gradle
+++ b/archaius2-api/build.gradle
@@ -19,6 +19,7 @@ apply plugin: 'java-library'
 dependencies {
     api     'javax.inject:javax.inject:1'
     implementation     'org.slf4j:slf4j-api:1.7.36'
+    testImplementation 'junit:junit:4.13.2'
 }
 
 eclipse {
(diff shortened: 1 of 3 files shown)


kilink:
See JDK implementation of these methods here: https://github.com/openjdk/jdk/blob/d10f277bd39bb5ac9bd48939c916de607fef8ace/src/java.base/share/classes/sun/reflect/generics/reflectiveObjects/ParameterizedTypeImpl.java

kilink:
Also note the Javadoc for [ParameterizedType](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/ParameterizedType.html), which states:

> Instances of classes that implement this interface must implement an equals() method that equates any two instances that share the same generic type declaration and have equal type parameters.

rgallardo-netflix (approved):

PR #705, opened 2024-02-25
Simplify KeyAndType equals / hashCode

kilink:
Simplify equals / hashCode for KeyAndType, add toString implementation for debugging. Additionally, avoid Optional / lambda allocation in some places.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/DefaultPropertyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/DefaultPropertyFactory.java
@@ -13,7 +13,7 @@ import java.lang.reflect.Type;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.List;
-import java.util.Optional;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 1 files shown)


archaius2-core/src/main/java/com/netflix/archaius/DefaultPropertyFactory.java
```diff
@@ -321,46 +326,48 @@ public KeyAndType(String key, Type type) {
...
 
         @Override
         public int hashCode() {
-            final int prime = 31;
             int result = 1;
-            result = prime * result + ((key == null) ? 0 : key.hashCode());
-            result = prime * result + ((type == null) ? 0 : type.hashCode());
+            result = 31 * result + Objects.hashCode(key);
+            result = 31 * result + Objects.hashCode(type);
             return result;
```
rgallardo-netflix, line 350: Couldn't we just return `Objects.hash(key, type)`?

rgallardo-netflix (approved):

archaius2-core/src/main/java/com/netflix/archaius/DefaultPropertyFactory.java
kilink: I didn't use it because it's a vararg method, so allocates an array. Maybe it doesn't matter here, or would get inlined by the JIT anyway, I don't know.

PR #708, opened 2024-02-28
Add support for tracking individual properties' stack traces

akang31:
Surface builder handle `setPropertiesToTrack`, AccessMonitorUtil handle `setPropertiesToTrack`, and `getTrackedPropertyTraces` to allow for tracking stack traces of specific properties and also pulling of the data when necessary.

We allow updating this on the fly for better fast property support

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
@@ -9,8 +9,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
```diff
@@ -28,6 +30,10 @@ public class AccessMonitorUtil implements AutoCloseable {
 
     // Map from stack trace to how many times that stack trace appeared
     private final ConcurrentHashMap<String, Integer> stackTrace;
+    // Property keys that we will keep the stack traces for
+    private Set<String> propertiesToTrack;
```
rgallardo-netflix: If the intent is that this set be settable on the fly, make it `volatile`.

rgallardo-netflix (changes requested):

archaius2-core/src/main/java/com/netflix/archaius/instrumentation/AccessMonitorUtil.java
akang31: Changed to be volatile

rgallardo-netflix (approved):

PR #724, opened 2024-05-21
Small fixes for collection types in configuration proxies:

### LINKED ISSUE #491: ConfigProxyFactory support @DefaultValue for Collections

### LINKED ISSUE #681: Add support for returning collections in parameterized properties
While normal properties support returning collections via the ConfigProxyFactory, parameterized properties do not.

e.g.
```
@Configuration
interface Foo {
  @PropertyName(name = "foo.bar.${0}")
  Set<String> getFooBarSet(String s);
}
```

rgallardo-netflix:
* Enable usage of the @DefaultValue annotation for methods with collection return types. Fixes: #491
* Enable usage of collection types in parametrized properties. Fixes: #681


akang31 (approved):
Looks great, thanks for the thorough tests!

PR #734, opened 2024-09-25
Improve handling of error messages on decoding errors.

rgallardo-netflix:
Ensure that all code paths that can throw a type error will log it properly and with as much information as possible.

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/DefaultPropertyFactory.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/DefaultPropertyFactory.java
@@ -28,8 +28,7 @@ public class DefaultPropertyFactory implements PropertyFactory, ConfigListener {
     
     /**
      * Create a Property factory that is attached to a specific config
-     * @param config
-     * @return
+     * @param config The source of configuration for this factory.
      */
     public static DefaultPropertyFactory from(final Config config) {
         return new DefaultPropertyFactory(config);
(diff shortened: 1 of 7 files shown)


kilink (approved):

archaius2-core/src/main/java/com/netflix/archaius/DefaultPropertyFactory.java
```diff
@@ -130,7 +131,7 @@ public <T> Property<T> asType(Function<String, T> mapper, String defaultValue) {
                         try {
                             return mapper.apply(value);
                         } catch (Exception e) {
-                            LOG.warn("Invalid value '{}' for property '{}'", propName, value);
+                            LOG.error("Invalid value '{}' for property '{}'. Will return the default instead.", propName, value);
```
akang31, line 134: I think due to the implementation of the above asType method, this just ends up returning null (this is what we observed in https://netflix.slack.com/archives/C141EJAJU/p1723751704356699)

rgallardo-netflix: Yeah, I saw that. It's the same as in the `get()` method. I'll delve into fixing those two and any others I find in a separate PR, I want that one to be as clean as I can so it can be called out in the release notes and also it's easy to pinpoint if/when people complain about a behavior change.

akang31: OK LGTM if this is assuming we are changing to that behavior, though this error message is technically a little off until we get to that point

akang31 (approved):

PR #735, opened 2024-09-25
Throw ParseException when a configuration setting can't be parsed as the type requested

rgallardo-netflix:
Homogenize the handling of parsing errors. Before this change, some code paths in the default Property implementation would silently swallow parsing errors and either return a defaultValue or null (even for non-nullable methods!) This leads to errors that are hard to debug.

Throwing a ParseException is a behavior change that will simplify debugging at the expense of surfacing latent bugs in user code. For example, if bad values are being set in config files, the code will now fail instead of silently returning a different value than expected.


kilink (approved):

akang31 (approved):
LGTM! 

Updated PropertyImpl looks good [x]
Error log is updated to indicate new behavior [x]
Test for the Property.orElse case [x]

PR #739, opened 2024-11-13
Add support for parsing Collections and Maps from the Spring YML format

akang31:
Add support for understanding the Spring YML (YamlPropertiesFactoryBean) way of parsing YML files via the Archaius endpoints. This supports the ConfigProxy and PropertyRepository as these all are eventually backed by AbstractConfig. 

This enables all existing Archaius read methods to read multi-line/multi-property maps and arrays as specified by Spring.

Spring YML files are converted to configs from this yml file
```yml
test:
    map:
        key1: 1
        key2: 2
        key3: 3
    stringList:
        - a
        - b
        - c
```

To these configs:
```
test.map.key1=1
test.map.key2=2
test.map.key3=3
test.stringList[0]=a
test.stringList[1]=b
test.stringList[2]=c
```

### DIFF
--- a/archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
@@ -3,6 +3,7 @@ package com.netflix.archaius.api;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -79,6 +80,14 @@ public final class ArchaiusType implements ParameterizedType {
     @Override
     public Type getRawType() {
         return rawType;
+    }
+
+    public boolean isMap() {
+        return Map.class.isAssignableFrom(rawType);
+    }
+
+    public boolean isCollection() {
+        return Collection.class.isAssignableFrom(rawType);
     }
 
     @Override
--- a/archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
@@ -3,7 +3,6 @@ package com.netflix.archaius.api;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
```diff
@@ -81,6 +82,14 @@ public Type getRawType() {
         return rawType;
     }
 
+    public boolean isMap() {
```
rgallardo-netflix: Maybe move these two to be private static helper methods in AbstractConfig? Then things would work with any `ParametrizedType`, not just with `ArchaiusType`.

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
```diff
@@ -298,6 +299,29 @@ protected <T> T getValue(Type type, String key) {
     @SuppressWarnings("unchecked")
     protected <T> T getValueWithDefault(Type type, String key, T defaultValue) {
         Object rawProp = getRawProperty(key);
+        if (rawProp == null && type instanceof ArchaiusType) {
```
rgallardo-netflix: I don't ... hate this, but let's think about it a bit more. I'd really love it if there was some way we could call the existing decoders here instead of marshaling the entire map into a single string just to unmarshal it back a nanosecond later. The same goes for the List support.

archaius2-api/src/main/java/com/netflix/archaius/api/ArchaiusType.java
akang31: Good call, moved these into the AbstractConfig instead

archaius2-core/src/main/java/com/netflix/archaius/config/AbstractConfig.java
akang31: I updated the code to instead rely on the decoder inside the AbstractConfig. The typing does become a little more annoying, and I also didn't include some of the DefaultCollectionsTypeConverterFactory types (specifically SortedSet and LinkedList). Honestly no great reason for that, but our SBN docs only specifically reference Map, List, and Set.

rgallardo-netflix (changes requested):
I think this is almost there!
Could you add tests for ... ?
* Reading these through a config proxy
* Reading a non-string yml value (ie, something like `key:  1` with an unquoted int)
* No regressions when reading the raw `mapNamePrefix.mapKey` as a "normal" key

akang31:
> I think this is almost there! Could you add tests for ... ?
> 
> * Reading these through a config proxy
> * Reading a non-string yml value (ie, something like `key:  1` with an unquoted int)
> * No regressions when reading the raw `mapNamePrefix.mapKey` as a "normal" key

Added some more tests to address these cases

rgallardo-netflix (approved):

Issue #644, opened 2023-03-24
Defend against empty strings in URLConfigReader constructor


rgallardo-netflix:
An equivalent issue was reported by @liuchong against 1.x as https://github.com/Netflix/archaius/pull/620

rgallardo-netflix:
Fixed by #741

PR #741, opened 2025-01-17
filter out empty or null string while parsing urls

tulikabhatt:
[Issue 664](https://github.com/Netflix/archaius/issues/644): Filtering out empty or null string before creating the URLs array

### DIFF
--- a/archaius2-core/src/main/java/com/netflix/archaius/readers/URLConfigReader.java
+++ b/archaius2-core/src/main/java/com/netflix/archaius/readers/URLConfigReader.java
@@ -18,6 +18,7 @@ package com.netflix.archaius.readers;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


archaius2-core/src/main/java/com/netflix/archaius/readers/URLConfigReader.java
```diff
@@ -56,10 +56,13 @@ private static URL[] createUrls(String... urlStrings) {
         if (urlStrings == null || urlStrings.length == 0) {
             throw new IllegalArgumentException("urlStrings is null or empty");
         }
-        URL[] urls = new URL[urlStrings.length];
+        final String[] validUrlStrings = Arrays.stream(urlStrings)
+                .filter(s -> s != null && !s.isEmpty())
+                .toArray(String[]::new);
```
rgallardo-netflix: Since you're already using a `Stream`, you might as well bring the `new Url(s)` to it. `Arrays.stream().filter().map().toArray(Url[])`. Avoid allocating an intermediate `String[]`.

rgallardo-netflix (approved):

PR #743, opened 2025-02-13
Swallow exceptions when calling toString on ConfigProxy objects

akang31:
We are currently running into an issue where a user has extended the ConfigProxy interface in a custom manner, adding support for, effectively, Map<String, JSONObject>. We do not support this kind of map (as our decoders assume that a single value is decoded into values, instead of collapsing a tree into a single object).

We want to avoid unnecessary logging for this case, so specifically for the toString call on ConfigProxy objects, we choose to swallow exceptions.

### DIFF
--- a/archaius2-api/src/main/java/com/netflix/archaius/api/Property.java
+++ b/archaius2-api/src/main/java/com/netflix/archaius/api/Property.java
@@ -60,15 +60,6 @@ public interface Property<T> extends Supplier<T> {
     T get();
 
     /**
-     * Return the most recent value of the property, swallowing any exceptions that occur.
-     *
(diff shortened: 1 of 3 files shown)


nowucca (approved):
I don't mind swallow as a word here, but could also use Ignores.
Up to you.

invokeIgnoringErrors
IgnoresErrorsProperty
getIgnoresErrors

akang31:
> I don't mind swallow as a word here, but could also use Ignores. Up to you.
> 
> invokeIgnoringErrors IgnoresErrorsProperty getIgnoresErrors

I like ignore as well here, switched to ignore

rgallardo-netflix (approved):

PR #765, opened 2026-05-30
reject malformed map entries in collections type converter

sahvx655-wq:
createMapTypeConverter in DefaultCollectionsTypeConverterFactory splits each comma-separated pair on '=' and reads kv[1] without checking the split produced two parts, so a Map property value like `key=12345,bogus` throws ArrayIndexOutOfBoundsException from inside the converter. Reject the malformed entry with a ParseException instead, matching how convertBoolean already rejects unparseable input.


sahvx655-wq:
gentle ping

PR #767, opened 2026-06-03
make listDelimiter volatile

sahvx655-wq:
reading the change that made decoder and interpolator volatile, listDelimiter is the third setter-backed field in AbstractConfig with the same shape: setListDelimiter writes it and getListDelimiter reads it (CommonsToConfig joins on it from the read path). it was missed, so a write still has no happens-before with reads on other threads and a stale delimiter can be observed indefinitely. same single-assignment setter, so volatile is the right level here too.


sahvx655-wq:
any update?

